PR #11, opened 2021-03-17
build: include proprietary CSS in Storybook

Robbert:
This configuration change fixes the color demo's in the Storybooks color page:

Screenshot 2021-03-17 at 17 19 11

Typical case of "worked in dev"... `npm run storybook` was okay, `npm run build && npm run start` was not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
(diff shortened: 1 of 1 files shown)


Robbert:
Hmm... after deploying this PR by merging it to `main`, the Storybook is still not yet fixed. The `.css` files from the static directories are still `404 Page Not Found`. Needs more investigation.

PR #62, opened 2021-05-21
feat: component library package

Robbert:


proprietary/design-tokens/src/index.scss
```diff
@@ -0,0 +1,4 @@
+@import "brand";
```
Yolijn, line 1: Please align with `component/_index.scss` and use `./brand` instead

src/component-library/src/components.scss
```diff
@@ -0,0 +1,18 @@
+@import "../../blockquote/index";
```
Yolijn, line 1: Please add a comment why this is a good temporary solution and what it will look like in the near future ☺️

Yolijn (changes requested):

PR #104, opened 2021-06-16
docs: button component design token documentation

Robbert:
Add a table with all the component design tokens.

Screenshot 2021-06-16 at 13 28 25


Robbert:
We're likely going to build our own design token table rendering instead.

PR #272, opened 2021-08-03
docs: release strategy

Robbert:


bddjong (approved):

bddjong:
Just for clarity: The Hague doesn't actually depend on the typography packages yet. Currently still using a 1:1 copy of the [heading](https://github.com/nl-design-system/denhaag/blob/master/src/components/Typography/src/heading.css) and [paragraph](https://github.com/nl-design-system/denhaag/blob/master/src/components/Typography/src/paragraph.css) css, as the packages weren't on npm yet when we created those components. Created an issue in our repo to update that https://github.com/nl-design-system/denhaag/issues/352

PR #320, opened 2021-08-16
build: update configuration for `@percy/storybook@4.0`

Robbert:


Robbert:
We're no longer using Percy

PR #352, opened 2021-08-25
docs: use correct orientation terms for spacing

bddjong:
Stumbled across a very confusing mixup of these two words while reading the docs for spacing tokens


bddjong:
Workflow cancelled for the sake of percy

PR #463, opened 2021-10-20
docs: moved Select component documentation out of Figma

usethetics:
Moved Select component documentation out of Figma


Robbert:
New docs are in Dutch

PR #501, opened 2021-11-13
fix: patch Stencil to copy nonce for <style> elements

Robbert:
In order to support `Content-Security-Policy` without `unsafe-inline`, the `<style>` elements from Stencil must contain a `nonce` attribute. The nonce value must be generated randomly for every page request. In order to for Stencil to know what nonce to use, this patch will make Stencil look for any existing `<style nonce="123...">` attributes, and copy the value.

In order to use the Web Components with CSP (and CSP is always a good idea) include the following in the `<head>` of your page:

```html
<html>
  <head>
    <style nonce="RANDOMLY-GENERATE-THIS-STYLE-NONCE"></style>
    <script nonce="RANDOMLY-GENERATE-THIS-SCRIPT-NONCE"></script>
```

You can use empty tags like the example above, but a nonce on any `style` element will work as long as it is available before the web component code is loaded.

The nonce must match the value from the header, for example:

```text
Content-Security-Policy: default-src 'self'; script-src 'self' 'nonce-RANDOMLY-GENERATE-THIS-SCRIPT-NONCE'; style-src 'self' 'nonce-RANDOMLY-GENERATE-THIS-STYLE-NONCE';
```

### DIFF
--- a/.patches/@stencil+core+2.9.0.patch
+++ b/.patches/@stencil+core+2.9.0.patch
@@ -1,0 +1,27 @@
+diff --git a/node_modules/@stencil/core/internal/client/index.js b/node_modules/@stencil/core/internal/client/index.js
+index eed9f0a..14b8f7b 100644
+--- a/node_modules/@stencil/core/internal/client/index.js
++++ b/node_modules/@stencil/core/internal/client/index.js
+@@ -275,6 +275,11 @@ const addStyle = (styleContainerNode, cmpMeta, mode, hostElm) => {
+                     if (BUILD.hydrateServerSide || BUILD.hotModuleReplacement) {
+                         styleElm.setAttribute(HYDRATED_STYLE_ID, scopeId);
+                     }
++                    const nonceElm = styleContainerNode.ownerDocument.querySelector('style[nonce]');
++                    if (nonceElm) {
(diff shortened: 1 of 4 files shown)


Robbert:
I used the following page for testing this:

```html
<!DOCTYPE html>
<html lang="en" class="utrecht-theme">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style nonce="GENERATE-THIS-STYLE-NONCE"></style>
    <script nonce="GENERATE-THIS-SCRIPT-NONCE"></script>
    <meta
      http-equiv="Content-Security-Policy"
      content="default-src 'self'; font-src 'self' data:; form-action 'self' https:; frame-src 'self'; img-src 'self' data:; object-src 'none'; script-src 'self' 'nonce-GENERATE-THIS-SCRIPT-NONCE'; style-src 'self' 'nonce-GENERATE-THIS-STYLE-NONCE';"
    />
    <title>Example of web components - Utrecht Design System</title>
    <script src="dist/utrecht/utrecht.esm.js" type="module"></script>
  </head>
  <body>
    <utrecht-logo></utrecht-logo>
  </body>
</html>
```

The developer console should display no errors related to CSP.

PR #524, opened 2021-11-25
feat: text-decoration-thickness design tokens for Link

Robbert:
`a:hover` in a paragraph:

Screenshot 2021-11-25 at 20 57 07

some findings:

## `text-decoration-skip`

`text-decoration-skip` can be helpful to avoid making some texts unreadable.
For example by obscuring Arabic diacritics.

However, the combination of a greater thickness and skipping this thick underline
leads to a very unappealing rendering of the underline. To avoid this,
`text-decoration-skip` is disabled for interactive states.

For design token configurations that have identical thickness for normal and interactive
states, this will lead to the (undesirable) effect that the focus/hover effect is switching
from an interrupted to an uninterrupted underline (for some texts).

Apart from making `skip-ink` configurable for normal/focus/hover, there is no good solution yet.

---

Disabling `text-decoration-skip` for interactive states obscures some texts, and we assume for now
that moving the pointer away from a link or having focus elsewhere first is simple enough to
not make this too inconvenient.

---

Some folks implement the underline of links using `border-bottom` or even using a finely crafted
`linear-gradient()` with a solid color at the bottom and transparent behind the text. These approaches
would unfortunately not be able to provide the improved readability of `text-decoration-skip`.

## `text-decoration-thickness`

Varying `text-decoration-thickness` can be used to distinguish interactive states.

---

`text-decoration-thickness` appears to have rendering differences between Chrome and Safari.
In Safari the line becomes thicker with extra pixels added to the bottom, while in Chrome
the underline offset also seems to increase along with the thickness, which effectively means
the underline is closer to the next line than in Safari.

---

It might be nice to use font-relative units for `text-decoration-thickness`, and that is why we
use the `max()` function to ensure the underline remains visible for every font size.

## `transition`

`text-decoration-thickness` could be a candidate for animating between interactive states,
however browsers don't seem to have implemented great looking supixel tweening yet.

`text-decoration-skip` also makes the transition more challenging to implement.

### DIFF
--- a/components/link/bem.scss
+++ b/components/link/bem.scss
@@ -4,12 +4,72 @@
  * Copyright (c) 2021 Gemeente Utrecht
  */
 
+// Avoid false positive for new CSS function `max()`
+/* stylelint-disable scss/no-global-function-names */
+
 @import "../common/focus/bem";
+
+/*
+
+# CSS implementation
+
+## `text-decoration-skip`
+
+`text-decoration-skip` can be helpful to avoid making some texts unreadable.
+For example by obscuring Arabic diacritics.
+
+However, the combination of a greater thickness and skipping this thick underline
+leads to a very unappealing rendering of the underline. To avoid this,
+`text-decoration-skip` is disabled for interactive states.
+
+For design token configurations that have identical thickness for normal and interactive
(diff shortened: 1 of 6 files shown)


JeroenduC:
Wat mij betreft helemaal top.

PR #649, opened 2022-01-05
refactor: omit .value in token reference

Robbert:
Let's get rid of the most annoying thing of the design token JSONs: the .value at the end of references. Omitting .value is supported since style-dictionary@3.1.0.

Tested by starting and smoke testing Storybook.


Yolijn (approved):
LGTY

- [x] Diff between dist output of `main` branch and this branch is only the date 👍

PR #650, opened 2022-01-05
Update major versions

Yolijn:
- [x] Add dev dependency of post-css to match required peer dependency of stylelint@14
- [x] `npm run update-major`
- [x] `npm run test-update`
- [x] `npm run storybook` smoke test

### DIFF
--- a/.ncurc.major.js
+++ b/.ncurc.major.js
@@ -26,5 +26,7 @@ module.exports = {
     'sass-loader',
     'style-loader',
     'webpack',
+    // We don't yet use Node 17 so the types of node should also be locked at Node 16.
+    '@types/node',
   ],
 };
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,9 @@ updates:
     reviewers:
       - "nl-design-system/sysadmin"
     ignore:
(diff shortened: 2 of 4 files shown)


packages/web-component-library-react/package.json
```diff
@@ -29,7 +29,7 @@
     "dist/"
   ],
   "devDependencies": {
-    "@types/node": "16.11.19",
+    "@types/node": "17.0.8",
```
Robbert, line 32: Let's include this package in the `npm-check-updates`/`dependabot.yml`, since Node.js is upgraded manually and we aim to keep the Node.js types in sync with the actual Node.js version. Incorrect version here could lead to errors about us using APIs that have changed in an upcoming version.

.ncurc.major.js
```diff
@@ -26,5 +26,7 @@ module.exports = {
     'sass-loader',
     'style-loader',
     'webpack',
+    // We don't yet use Node 17 so the types of node should also be locked at 6.
```
Robbert, line 29: "locked at 6" → "locked at Node 16"

Robbert (changes requested):

Robbert (approved):

PR #693, opened 2022-01-26
refactor: simplify BEM structure of sidenav

Robbert:
Screenshot 2022-01-26 at 16 14 58

Replaces ancient PR #295.


Ollie-nl (approved):
Time to see it in action

PR #700, opened 2022-02-04
Docusaurus heading component

Yolijn:
- [x] Add components to docusaurus
- [x] Improve heading and sidebars for docusaurus
- [x] Separate bem/css stories to enable neat iframes for each story


standalone/docusaurus/src/components/HomepageFeatures.js
```diff
@@ -1,11 +1,14 @@
+import Bicycles from '@site/static/img/bicycles.png';
+import Laptop from '@site/static/img/laptop.png';
+import Utrecht from '@site/static/img/utrecht.png';
```
Yolijn, line 3: By importing images like this, it doesn't break when the BaseUrl is changed at any point

.eslintignore
```diff
@@ -1,9 +1,11 @@
 # Ignore 3rd party files
 node_modules
+iframeResizer.contentWindow.min.js
```
Robbert, line 3: We could consider including `*.min.js` in "generated files"

standalone/docusaurus/docs/componenten/Heading.mdx
```diff
@@ -0,0 +1,88 @@
...
+  - h2
+  - h3
+  - h4
+  - h5
+  - h6
+  - heading
+  - utrecht-heading
+---
+
+import { Markdown } from "../../src/components/Markdown";
```
Robbert, line 20: I have tried to include `<!-- @license CC0-1.0 -->` in every `*.md`/`*.mdx` documentation page, so it easy to copy to other repos while maintaining the open source license and it is clear that the license for docs is different from the code repository overall.

components/heading-4/bem.stories.mdx
```diff
@@ -0,0 +1,60 @@
+<!-- @license CC0-1.0 -->
+
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { ComponentTokensSection } from "../../documentation/components/ComponentTokensSection";
+import tokens from "../../proprietary/design-tokens/dist/index.json";
+import "../heading-4/bem.scss";
```
Robbert, line 6: I kind of expected the import order ESLint plugin to complain that the `.css` import (side effect imports) should be the last.

Robbert (approved):

.eslintignore
Yolijn: As discussed, let's add another section for "Ignore minified files" as it's not generated by us and it's probably better not to make future developers assume something different :)

components/heading-4/bem.stories.mdx
Yolijn: The import order in mdx files isn't checked by our linting. But I've moved those side-effect imports to be the last import of the heading mdx files

PR #721, opened 2022-02-14
feat: add local search

Yolijn:
While we're waiting for the nice people from Algolia to grand us free access to their crawler services, let's try one of the local search packages from https://docusaurus.io/docs/search#using-local-search

- [x] Install one of the most used packages [docusaurus-lunr-search](https://www.npmjs.com/package/docusaurus-lunr-search)
- [x] Follow the steps to install it and use their search bar

🎉 Tada: local search works after build.
**Note**: it doesn't work during development with `npm run docs` but it does work out of the box after deployment and with `npm run build:docs && npm run serve:docs` locally

https://user-images.githubusercontent.com/877246/153872649-73af896c-98e0-493f-b043-04ad877e673a.mp4

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 # Ignore 3rd party files
 node_modules
+SearchBar
 
 # Ignore minified files
 *.min.js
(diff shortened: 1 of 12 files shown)


Robbert (approved):
lgtm

PR #779, opened 2022-03-18
fix: rename `table-heading` to `table-header-cell`

### LINKED ISSUE #601: Table CSS classes names
The BEM classes for the table include `utrecht-table__heading` and `utrecht-table__header`. Looking at the [HTML CSS](https://github.com/nl-design-system/utrecht/blob/main/components/table/html.scss#L19-L29), I believe the naming of these classes is very confusing.

HTML spec refers to `th` as being a **header** cell. But the CSS is using the `__heading` element. Whilst the `thead` selector is using the `__header` class.

Robbert:
Resolves #601

### DIFF
--- a/components/table/html.scss
+++ b/components/table/html.scss
@@ -25,7 +25,7 @@
 }
 
 .utrecht-html th {
-  @extend .utrecht-table__heading;
+  @extend .utrecht-table__header-cell;
 }
 
 .utrecht-html th,
--- a/components/table/bem.scss
+++ b/components/table/bem.scss
@@ -45,11 +45,11 @@
   vertical-align: baseline;
 }
 
-.utrecht-table__heading {
-  color: var(--utrecht-table-heading-color);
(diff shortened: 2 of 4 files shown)


components/table/html.scss
```diff
@@ -25,7 +25,7 @@
 }
 
 .utrecht-html th {
-  @extend .utrecht-table__heading;
+  @extend .utrecht-table__header-cell;
```
bddjong, line 28: What's the reasoning for adding the `-cell` part? `<th>` already indicates a single header. This kind of feels as though there is also `table-header` component. (Asking because we went for `table-header` with [Den Haag](https://github.com/nl-design-system/denhaag/blob/main/components/Table/src/TableHeader.tsx)).

Robbert: I went with `table__header`/`table__body`/`table__footer` for the `thead`/`tbody`/`tfoot` sections, so `header` was indeed already taken.

@Yolijn any thoughts?

PR #852, opened 2022-04-21
feat(button): border-color design token

Robbert:
Turns out we had the border-width token, but no way to configure the color of the border. 

Gemeente Buren had no way to correctly to correctly style their secondary button, this should fix that.

### DIFF
--- a/components/button/bem.scss
+++ b/components/button/bem.scss
@@ -7,6 +7,7 @@
 
 .utrecht-button {
   background-color: var(--utrecht-button-primary-action-background-color, var(--utrecht-button-background-color));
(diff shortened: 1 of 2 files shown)


Yolijn (approved):

Yolijn (changes requested):
☑️ Done

I think we need to at least also add `border-color` token for hover state. 
```
utrecht-button--hover:not(:disabled),
.utrecht-button:hover:not(:disabled):not([aria-disabled="true"]):not(.utrecht-button--disabled) {
...
border-color: var(--utrecht-button-hover-border-color, var(--utrecht-button-border-color));
...
}
```

PR #936, opened 2022-05-17
docs: add HTML panel for Storybook

Robbert:


Yolijn (approved):
Works on my machine too!

PR #1048, opened 2022-07-11
fix: angular packages

AliKdhim87:


.eslintignore
```diff
@@ -13,15 +13,14 @@ tmp
 
 # Ignore generated files from Stencil
 components/components.d.ts
-packages/web-component-library-angular/src/directives/angular-component-lib/utils.ts
-packages/web-component-library-angular/src/directives/boolean-value-accessor.ts
-packages/web-component-library-angular/src/directives/number-value-accessor.ts
+packages/*/dist
```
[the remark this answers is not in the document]
AliKdhim87: You are right, there is no need for it, I will remove it

packages/component-library-angular/src/table-header-cell/component.ts
```diff
@@ -12,6 +12,6 @@ import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@a
   },
 })
 export class UtrechtTableHeaderCellAttr {
-  @Input() scope: string = null;
+  @Input() scope?: string;
```
Robbert, line 15: I had used `scope = null` instead of `scope?` because in the DOM for Elements usually have properties taht are `null` if they are not set instead of `undefined`, and I meant to mimic that behavior for our custom elements.

packages/web-component-library-angular/package.json
```diff
@@ -22,28 +22,33 @@
     "clean": "rimraf dist/"
   },
   "peerDependencies": {
-    "rxjs": "^6.6.0"
+    "@angular/common": "12 - 14",
+    "@angular/core": "12 - 14"
```
Robbert, line 26: Since Angular seems to be so great at backwards compatibility, I was thinking to use `>=11` instead of specifically limiting new versions beyond 14.

packages/web-component-library-react/package.json
```diff
@@ -33,16 +33,16 @@
...
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
     "rimraf": "3.0.2",
     "typescript": "4.3.5"
   },
   "peerDependencies": {
-    "react": "^16.7.0 || ^17.0.0",
-    "react-dom": "^16.7.0 || ^17.0.0"
+    "react": "^16 - ^18",
+    "react-dom": "^16 - ^18"
```
Robbert, line 45: Same here, what would you think of `>=16`?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compileOnSave": false,
+  "compilerOptions": {
+    "baseUrl": "./",
```
Robbert, line 4: I prefer we always use relative paths to make it super clear what an import is referring to. What would you think of not defining `baseUrl` here? Even if you would want to use it in packages, I think it would be confusing that in a package that extends this `tsconfig` the baseUrl would be relative to the root of this repo.

packages/component-library-angular/src/table-header-cell/component.ts
AliKdhim87: I changed to `scope?:` because the TypeScript was complaining about  `Type 'null' is not assignable to type 'string'.ts(2322)`  but I can change ito `@Input() scope: string | null` ?

PR #1061, opened 2022-07-18
docs: add storybook vue

AliKdhim87:
>Note: There is an issue with the storybook show code button, it shows not the correct code snippet it's a storybook [bug](https://github.com/storybookjs/storybo
Screenshot 2022-07-18 at 17 35 05
ok/issues/13917), I tried to implement this [solution ](https://github.com/storybookjs/storybook/issues/13917#issuecomment-1013019386), but it didn't work, 
 I added this as a temporary solution until we find something different 
Screenshot 2022-07-18 at 17 35 30


packages/storybook-vue/src/stories/Introduction.stories.mdx
```diff
@@ -0,0 +1,8 @@
+import { Description, Meta } from "@storybook/addon-docs";
+import README from "../../../component-library-vue/README.md";
```
Robbert, line 2: Probably nicer to import from `@utrecht/component-library-vue`

Robbert (approved):

PR #1066, opened 2022-07-19
docs: add button states

AliKdhim87:


packages/storybook-vue/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@utrecht/component-library-vue": "file:../component-library-vue",
     "@utrecht/design-tokens": "file:../../proprietary/design-tokens",
+    "storybook-addon-pseudo-states": "1.15.1",
     "vue": "3.2.37"
```
Robbert, line 19: Shouldn't all these dependencies be a `devDependency`? After `npm run build` everything is contained in `dist/` and it should have no further dependencies.

PR #1086, opened 2022-07-25
docs: add vue link stories

AliKdhim87:


packages/storybook-vue/src/stories/Link/Link.stories.mdx
```diff
@@ -0,0 +1,187 @@
...
+          code: `<Link :external="true" target="_blank" rel="noopener">External</Link>`,
+        },
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+## Eamil
```
Robbert, line 51: Eamil → Email

packages/storybook-vue/src/stories/Link/Link.stories.mdx
```diff
@@ -0,0 +1,187 @@
...
+          </Link>
+          </div>
+          `,
+        },
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
Robbert, line 185: I think we can remove the Arabic example from the Vue docs, since it was mainly included to tweak the CSS `text-underline-offset`, there is no special API or class name for this.

packages/storybook-vue/src/stories/Link/Link.stories.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+# Link
+
+export const Template = (args) => ({
+  components: { Link },
+  setup() {
+    return { args };
+  },
+  template: `<div dir="${args.dir}" lang="${args.lang}"><Link v-bind="args">${args.default}</Link></div>`,
+});
```
Robbert, line 15: All these stories also render a `default` attribute in HTML, as well as `dir` and `lang` attributes. Is this a bug?

Screenshot 2022-07-26 at 10 46 01

packages/storybook-vue/src/stories/Link/Link.stories.mdx
```diff
@@ -0,0 +1,187 @@
...
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+## External
+
+<Canvas>
+  <Story
+    name="External"
+    args={{ ...componentArgs, default: "Extenal component", external: true, target: "_blank", rel: "noopener" }}
```
Robbert, line 38: `Extenal component` → `External link`

packages/storybook-vue/src/stories/Link/argTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+    defaultValue: false,
+  },
+  href: {
+    name: 'href',
+    type: { name: 'string', required: false },
+    defaultValue: false,
+  },
+  target: {
+    type: 'select',
+    options: ['_self', '_blank', '_parent', '_top'],
```
Robbert, line 14: I'm not sure we need to include any `_self`, `_parent` and `_top`. Just empty string and `_blank` would be sufficient.

Including `_parent` and `_top` implies we are okay with using frames and and `<base target="_new">` and I don't think we should be.

packages/storybook-vue/src/stories/Link/argTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+};
+
+export const componentArgs = {
+  external: false,
+  href: 'https://example.com',
+  default: 'Default Link',
```
Robbert, line 21: What does `default` mean in this context? `textContent`?

packages/storybook-vue/src/stories/Link/Link.stories.mdx
AliKdhim87: Yes, I think this is a bug, here is the reference https://github.com/storybookjs/storybook/discussions/12691
I tried a couple of scenarios but it is still the same, also it's because these properties are not exist in the component itself we tried to add them through the storybook.
But, I don't see it as a big issue because it's only happening in storybook

packages/storybook-vue/src/stories/Link/argTypes.ts
AliKdhim87: `default` is a property defined in the storybook to refer to the `<slot>`, also it can be any name we want

AliKdhim87: I think the `textContent` name makes a lot of sense so I will update it!

Robbert (approved):

PR #1094, opened 2022-07-28
feat: refactor vue select component

AliKdhim87:


packages/component-library-vue/src/UtrechtSelect.vue
```diff
@@ -1,17 +1,28 @@
...
+        selected: useModelWrapper(props, emit),
+      };
+    },
+  });
 </script>
 
 <template>
   <select
+    v-if="options && options.length > 0"
+    v-model="selected"
```
Robbert, line 25: Do I understand correctly that this code removes the `<select>` when there are no options? I would expect it would just render the `<select>` with no children.

packages/component-library-vue/src/UtrechtSelect.vue
```diff
@@ -1,17 +1,28 @@
-<script setup lang="ts">
-  defineEmits<(_type: "update:modelValue", _value: unknown) => void>();
+<script lang="ts">
+  import { defineComponent } from "vue";
+  import { useModelWrapper } from "./helpers";
```
Robbert, line 3: Why not directly import from `./modelWrapper`? Seems like having two files is not necessary.

packages/component-library-vue/src/UtrechtSelect.vue
```diff
@@ -23,10 +34,18 @@
     :disabled="disabled"
     :readonly="readonly"
     :required="required"
-    :value="modelValue"
-    @input="$emit('update:modelValue', (($event as InputEvent).target as HTMLSelectElement).value)"
   >
-    <slot />
+    <option
+      v-for="{value, label} in options"
```
Robbert, line 39: I think one property is missing: `v-for="{disabled, value, label} in options"`

packages/component-library-vue/src/UtrechtSelect.vue
```diff
@@ -23,10 +34,18 @@
...
+      :key="value"
+      :value="value"
+      :class="[
+        {'utrecht-select__option--disabled': disabled},
```
Robbert, line 43: The `disabled` attribute and the `utrecht-select__option--disabled` class name should only be used when this specific option is disabled, not when the entire select is disabled.

packages/storybook-vue/src/stories/Select/SelectStates.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+# Select States
+
+export const Template = (args) => ({
+  components: { Select },
+  setup() {
+    return { args };
+  },
+  template: `
+  <Select v-bind="args" @change="action" />
```
Robbert, line 16: I was wondering if we should make the `update:modelValue` event also show up in the Actions panel, to demonstrate the form bindings. What do you think?

packages/storybook-vue/src/stories/Select/SelectStates.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+      docs: {
+        source: {
+          code: `<Select :options="options" />`,
+        },
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
Robbert, line 128: Seems like the one state missing is now `readonly`.

packages/storybook-vue/src/stories/Select/SelectWithFieldsetAndLegend.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+`,
+        },
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable of={Select} />
```
Robbert, line 64: For radio groups and checkbox groups I see the added value of demonstrating the fieldset, but for `<select>` I'm not so sure if we need it. What do you think?

packages/component-library-vue/src/UtrechtSelect.vue
```diff
@@ -1,17 +1,28 @@
...
-    required?: boolean;
-  }>();
+  export default defineComponent({
+    props: {
+      disabled: { type: Boolean, required: false },
+      invalid: { type: Boolean, required: false },
+      readonly: { type: Boolean, required: false },
+      required: { type: Boolean, required: false },
+      options: { type: [Array, Object], required: true },
+      modelValue: { type: [String, Number, Boolean], require: false, default: "" },
```
Robbert, line 12: I wonder if we should also invest some time, in the near future, to support generics for `modelValue` so we allow any type here. `defineComponent<T>(` or something, if you know what I mean...

Robbert (changes requested):

packages/component-library-vue/src/UtrechtSelect.vue
AliKdhim87: Yes, I think having `select` without children is better in case there are no options, but I got `Eslint` [warning](https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html) Vue preventing the use of ` v-for` directives together with `v-if` directives on the same element.  So, can we have a `.eslintrc` in the `/component-library-vue` to adjust some rules,

packages/component-library-vue/src/UtrechtSelect.vue
AliKdhim87: I was thinking about making it nicer  when you import the helper's functions from one path, but I will change it to your suggestion because currently, we have one function

packages/storybook-vue/src/stories/Select/SelectWithFieldsetAndLegend.stories.mdx
AliKdhim87: I think we do not need it, so I will update it

packages/component-library-vue/src/UtrechtSelect.vue
AliKdhim87: I Agree

packages/component-library-vue/tsconfig.lib.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "@vue/tsconfig/tsconfig.web.json",
-  "include": ["src/**/*", "src/**/*.vue"],
+  "include": ["src/**/*", "src/**/*.vue", "src/**/*ts"],
```
Robbert, line 3: `*ts` → `*.ts`

Robbert (approved):
One typo left, we'll fix it in the next PR.

PR #1105, opened 2022-08-01
docs: refactor heading with level stories

AliKdhim87:
the reason why because we need to add the another Heading component


packages/storybook-vue/src/stories/Heading/Heading.stories.mdx
```diff
@@ -10,93 +10,46 @@ export const Template = (args) => ({
...
   },
-  template: `<Heading v-bind="args">${args.default}</Heading>`,
+  template: `
+  <Heading v-bind="args" :level="args.level1">${args.headingContentLevel1}</Heading>
+  <Heading v-bind="args" :level="args.level2">${args.headingContentLevel2}</Heading>
+  <Heading v-bind="args" :level="args.level3">${args.headingContentLevel3}</Heading>
+  <Heading v-bind="args" :level="args.level4">${args.headingContentLevel4}</Heading>
+  <Heading v-bind="args" :level="args.level5">${args.headingContentLevel5}</Heading>
+  <Heading v-bind="args" :level="args.level6">${args.headingContentLevel6}</Heading>
+  `,
```
Robbert, line 20: Can't we keep it simple and hardcode the `:level="1"` etc?

packages/storybook-vue/src/stories/Heading/Heading.stories.mdx
```diff
@@ -10,93 +10,46 @@ export const Template = (args) => ({
...
+      level3: 3,
+      level4: 4,
+      level5: 5,
+      level6: 6,
+      headingContentLevel1: "The Quick Brown Fox Jumps Over The Lazy Dog",
+      headingContentLevel2: "The Quick Brown Fox Jumps Over The Lazy Dog",
+      headingContentLevel3: "The Quick Brown Fox Jumps Over The Lazy Dog",
+      headingContentLevel4: "The Quick Brown Fox Jumps Over The Lazy Dog",
+      headingContentLevel5: "The Quick Brown Fox Jumps Over The Lazy Dog",
+      headingContentLevel6: "The Quick Brown Fox Jumps Over The Lazy Dog",
```
Robbert, line 38: How about using only `textContent` property that is used for every header? You probably want to compare the same text in different fonts anyway.

packages/storybook-vue/src/stories/Heading/tokensHeadingLevel1.stories.mdx
```diff
@@ -0,0 +1,22 @@
...
+  parameters={{
+    status: {
+      type: "ALPHA",
+    },
+  }}
+/>
+
+# Heading Design Tokens
+
+## Heading 1
```
Robbert, line 20: We can simplify this as `# Heading 1 Design Tokens`, I reckon.

packages/storybook-vue/src/stories/Heading/HeadingLevel2.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+    }}
+    parameters={{
+      docs: { source: { code: `<Heading :level="2">The Quick Brown Fox Jumps Over The Lazy Dog</Heading>` } },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable of={Heading} />
```
Robbert, line 36: All the `ArgsTable`s don't seem to work for me... do they work for you?

Screenshot 2022-08-01 at 13 17 07

packages/storybook-vue/src/stories/Heading/Heading.stories.mdx
AliKdhim87: Honestly, I don't think we need a property here, hardcoded text is more than enough  because the story role is just to display all heading in one page, like I said in the previous comment we have the controls individually

Robbert (approved):

PR #1107, opened 2022-08-01
feat: improve the vue textarea component

AliKdhim87:


packages/storybook-vue/src/stories/Textarea/TextareaStates.stories.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+## Focus
+
+<Canvas>
+  <Story
+    name="Focus"
+    args={{
+      placeholder: "Focus Textarea",
+      name: "focus-textarea",
+      id: "focus-textarea",
```
Robbert, line 127: For all these stories I think it would be more helpful to have an actual `value` (quick brown fox?) instead of a placeholder, this way we cannot test `...-focus-color`.

I suggest we add one separate story that shows `placeholder`.

packages/storybook-vue/src/stories/Textarea/TextareaStates.stories.mdx
```diff
@@ -0,0 +1,165 @@
...
+  template: `
+  <FormField> 
+    <Textarea v-bind="args" @update:modelValue="action" :value="args.modelValue" />
+  <br/>
+  <br/>
+  <Paragraph v-if="args.modelValue"><strong> Textarea Binding:</strong> {{args.modelValue}}</Paragraph>
+  </FormField>
+  `,
+  methods: { action: action("modelValue") },
+});
```
Robbert, line 25: I suggest we move this value binding example from States to a separate page.

packages/storybook-vue/src/stories/Textarea/Textarea.stories.mdx
```diff
@@ -0,0 +1,47 @@
...
+  template: `
+  <FormField> 
+    <Textarea v-bind="args" @update:modelValue="action" :value="args.modelValue" />
+  <br/>
+  <br/>
+  <Paragraph v-if="args.modelValue"><strong> Textarea Binding:</strong> {{args.modelValue}}</Paragraph>
+  </FormField>
+  `,
+  methods: { action: action("modelValue") },
+});
```
Robbert, line 25: I suggest we keep the default story of every component quite simple, and include a separate page in Storybook that shows form bindings for form components.

packages/component-library-vue/src/UtrechtTextarea.vue
```diff
@@ -1,32 +1,34 @@
...
-      'utrecht-textarea--readonly': readonly,
-    }"
-    :disabled="disabled"
+    :class="[
+    { 'utrecht-textarea--invalid': invalid },
+    { 'utrecht-textarea--readonly': readonly },
+    'utrecht-textarea',
+    'utrecht-textarea--html-textarea'
+    ]"
     :readonly="readonly"
```
Robbert, line 30: Is `readonly` still necessary to implement explicity, or is it also covered by `utrecht-textarea--html-textarea`?

packages/storybook-vue/src/stories/Textarea/TextareaStates.stories.mdx
AliKdhim87: I think you didn't get the focus color because you didn't build the `component-library-vue`

Screenshot 2022-08-01 at 13 34 26

From my prospect, I think each story has a placeholder with descriptive value, it's better than just text as value if it's only the focus issue.
WDYT?

AliKdhim87: So, I will update all stories with a filled `modelValue`, the reason why is because we need to display the  `<Textarea />` font CSS option when the input's states changed!
Thanks, @Robbert  for the explanation

AliKdhim87: Only the  Textarea `required` story I will leave the `modelValue` empty to demonstrate the state

packages/storybook-vue/src/stories/Textarea/TextareaBinding.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+  template: `
+  <FormField> 
+    <Textarea v-bind="args" @update:modelValue="action" :value="args.modelValue" />
+  <br/>
+  <br/>
+  <Paragraph v-if="args.modelValue"><strong> Textarea Binding:</strong> {{args.modelValue}}</Paragraph>
+  </FormField>
+  `,
+  methods: { action: action("modelValue") },
+});
```
Robbert, line 25: The real-time binding doesn't seem to work for me, unfortunately. The next never updates, even though the action panel shows the change.

Screenshot 2022-08-02 at 13 01 34

packages/storybook-vue/src/stories/Textarea/TextareaBinding.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+  setup() {
+    return { args };
+  },
+  template: `
+  <FormField> 
+    <Textarea v-bind="args" @update:modelValue="action" :value="args.modelValue" />
+  <br/>
+  <br/>
+  <Paragraph v-if="args.modelValue"><strong> Textarea Binding:</strong> {{args.modelValue}}</Paragraph>
+  </FormField>
```
Robbert, line 22: I think there are ways to use more semantic HTML for displaying the form binding.

Instead of inside `FormField`, we can use something like this after `<FormField>`:

```
<dl v-if="args.modelValue">
  <div aria-live="polite" aria-atomic="true">
    <dt>Textarea Binding:</dt>
    <dd>{{args.modelValue}}</dd>
  </div>
</dl>
```

packages/storybook-vue/src/stories/Textarea/TextareaBinding.stories.mdx
AliKdhim87: Yes, that is correct, I tried to solve it by following this [solution](https://craigbaldwin.com/blog/updating-args-storybook-vue/) but it seems the binding will be not in sync with the `modelValue` field below in the controls.
Only the `modelValue` in the controls change the value
Screenshot 2022-08-02 at 14 17 23
e

Robbert: I suspect having it out of sync like this will be quite confusing to folks. How about we remove this story for now, and include it once we have a syncing solution?

Robbert (approved):

PR #1110, opened 2022-08-02
feat: export the vue table component

AliKdhim87:


packages/storybook-vue/src/stories/Table/Table.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+# Table component
+
+export const Template = (args) => ({
+  components: { Table },
+  setup() {
+    return { args };
+  },
+  template: `
+  <Table v-bind="args" aria-labelledby="53818ecc-f9a6-4a9d-9e63-12a5ebb89e31">
+  <TableCaption id="53818ecc-f9a6-4a9d-9e63-12a5ebb89e31"> Table Caption </TableCaption>
```
Robbert, line 16: `aria-labelledby` should not be necessary for `caption`, it should work automatically. Did you run into issues with the label in the accessibility tree?

The main use for `aria-labelledby` would be if you have a situation where `h1`-`h6` or `figcaption` is preferable to `caption`.

packages/storybook-vue/src/stories/Table/argTypes.ts
```diff
@@ -0,0 +1,7 @@
+export const argTypes = {
+  disabled: {
+    name: 'disabled',
+    type: { name: 'boolean', required: false },
+    defaultValue: false,
+  },
```
Robbert, line 6: What does a `disabled` table look like?

AliKdhim87: I will remove it!

PR #1116, opened 2022-08-03
build: improve the icon script

AliKdhim87:


components/icon/src/build.js
```diff
@@ -1,94 +1,97 @@
...
-        `${dir}/${formattedName}.stencil.tsx`,
-        component(fileName, svg, iconRapperComponent),
-        writeFileErrorHandler,
+const getIconData = async () => {
+  let svgData;
+
+  const files = await readSVGDir();
+  try {
+    if (files && files.length > 0) {
+      svgData = await Promise.all(
```
Robbert, line 33: Why is there a `Promise.all` here? I don't think I can find the promise here.

components/icon/src/build.js
```diff
@@ -1,94 +1,97 @@
...
+    }
+  } catch (error) {
+    console.error(error);
+  }
+  return svgData;
+};
 
-  if (!dirExistsSync(outputDir)) {
-    fs.mkdirSync(outputDir);
+const WriteComponentFile = async () => {
```
Robbert, line 46: Starting `WriteComponentFile` with a capital is quite unusual for a function, mostly we use that for classes.

components/icon/src/build.js
```diff
@@ -1,94 +1,97 @@
...
+        const formattedName = kebabCase(fileName);
+        const svg = await readFile(`${cwd}/${file}`, 'utf-8');
+
+        await mkdir(`${componentsPath}/${formattedName}`, { recursive: true });
+        await writeFile(
+          `${componentsPath}/${formattedName}/${formattedName}.stencil.tsx`,
+          component(fileName, svg, iconRapperComponent),
+          'utf-8',
+        );
+        await writeFile(`${componentsPath}/${formattedName}/${formattedName}.spec.tsx`, test(fileName), 'utf-8');
```
Robbert, line 62: In the future you could consider an approach where you don't `await` for every file to finish writing, but to start writing to a couple of files in parallel and just `await Promise.all(...)`

packages/storybook/stories/icon/icon-gallery.stories.mdx
```diff
@@ -108,7 +108,7 @@ import iconSet from "@utrecht/icon/dist/index.json";
 ## Icon Set
 
 <CustomIconGallery>
-  {iconSet.map(({ id }, index) => (
-    <CustomIconGalleryItem name={id} key={index} icon={<id />} />
+  {iconSet.map(({ id: Icon }, index) => (
+    <CustomIconGalleryItem name={Icon} key={index} icon={<Icon />} />
```
Robbert, line 112: Thanks for fixing this, I think I broke this a while ago and forgot about it.

components/icon/src/build.js
AliKdhim87: Here is the answer https://stackoverflow.com/questions/40140149/use-async-await-with-array-map

components/icon/src/build.js
AliKdhim87: You are right, it's a typo

PR #1119, opened 2022-08-03
feat: textbox support `[attribute="value"]` in css

AliKdhim87:
The reason to minimize using javascript to deal style

**reference to the issue:**
https://github.com/nl-design-system/utrecht/blob/main/packages/component-library-react/src/Textbox.tsx#L33

### DIFF
--- a/components/textbox/css/index.scss
+++ b/components/textbox/css/index.scss
@@ -49,4 +49,21 @@
 
 .utrecht-textbox--html-input {
   @include utrecht-textbox--html-input;
+
+  &[type="password" i] {
+    @include utrecht-textbox--password;
+  }
+  &[type="url" i],
+  &[type="email" i],
+  &[inputMode="email" i],
+  &[inputMode="url" i] {
+    @include utrecht-textbox--url;
+  }
+  &[type="number" i],
+  &[type="tel" i],
+  &[inputMode="numeric" i],
+  &[inputMode="decimal" i],
+  &[inputMode="tel" i] {
+    @include utrecht-textbox--numeric;
+  }
 }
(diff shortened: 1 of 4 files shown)


components/textbox/css/_mixin.scss
```diff
@@ -138,6 +138,18 @@
...
 @mixin utrecht-textbox--html-input {
+  &[type="password" i] {
+    @include utrecht-textbox--password;
+  }
+  &[type="url" i],
+  [type="email" i] {
+    @include utrecht-textbox--url;
+  }
+  &[type="number" i],
+  [type="tel" i] {
```
Robbert, line 149: I think for `tel` and `email` the `&` prefix is missing in the selector.

And I think we can include `&[inputMode="numeric" i]`, `&[inputMode="decimal" i]` and `&[inputMode="tel" i]` too.

components/textbox/css/_mixin.scss
```diff
@@ -138,6 +138,18 @@
 }
 
 @mixin utrecht-textbox--html-input {
+  &[type="password" i] {
+    @include utrecht-textbox--password;
+  }
+  &[type="url" i],
+  [type="email" i] {
+    @include utrecht-textbox--url;
+  }
```
Robbert, line 147: We can include `&[inputMode="email" i]` and `&[inputMode="url" i]` too.

Robbert (commented):
I think now that we have these selectors here, the selectors generated in `textbox/html/index.css` will have duplicate attribute selectors, we might be able to address that in this PR too.

```css
input[type="week" i] {
  @include utrecht-textbox;
  @include utrecht-textbox--html-input;
}
```

And I think we can also clean up the React module to implement fewer class names.

PR #1122, opened 2022-08-04
build: use `main.js` to set the `staticDirs`

AliKdhim87:
https://storybook.js.org/docs/react/configure/images-and-assets#serving-static-files-via-storybook-configuration


Robbert (approved):
Much better this way, thanks!

PR #1126, opened 2022-08-05
feat: add vue custom radio button component

AliKdhim87:


packages/component-library-vue/src/CustomRadioButton.vue
```diff
@@ -0,0 +1,32 @@
...
+
+<template>
+  <input
+    v-model="value"
+    type="radio"
+    :aria-invalid="invalid || undefined"
+    :class="[
+        { 'utrecht-custom-radio-button--invalid': invalid },
+        'utrecht-custom-radio-button',
+        'utrecht-custom-radio-button--html-input'
```
Robbert, line 27: For consistency I do kinda wish we had a `utrecht-custom-radio-button--html-input` (to separate out the pseudo-class selectors), but we don't have one implemented yet.

I suggest we either implement `utrecht-custom-radio-button--html-input` or remove this class name for now.

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButton.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import type { Meta, StoryFn } from '@storybook/vue3';
+import { CustomRadioButton } from '@utrecht/component-library-vue';
+import { argTypes } from './argTypes';
+
+export default {
+  component: CustomRadioButton,
+  id: 'vue-custom-radio-button',
+  title: 'Vue.js Component/Custom Radio Button',
+  argTypes: argTypes,
```
Robbert, line 9: `argTypes: argTypes,` → `argTypes,`

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonStates.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export const Checked = Template.bind({});
+Checked.args = {
+  checked: true,
+  value: 'checked',
+};
+Checked.parameters = {
+  docs: {
+    source: {
+      code: `<CustomRadioButton :checked="true" value="checked"/>`,
```
Robbert, line 28: I wonder if these `value` attributes in the examples for CustomRadioButton might be more confusing than helpful. Especially this example, folks might think `value="checked"` is just as essential to copy as `:checked="true"`.

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonWithFieldsetAndLegend.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import { action } from '@storybook/addon-actions';
+import type { Meta, StoryFn } from '@storybook/vue3';
+import { CustomRadioButton } from '@utrecht/component-library-vue';
+import { argTypes, data } from './argTypes';
+export default {
+  component: CustomRadioButton,
+  id: 'vue-custom-radio-button-with-field-set-and-legend',
```
Robbert, line 7: `field-set` → `fieldset`

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonWithFieldsetAndLegend.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+const Template: StoryFn<typeof CustomRadioButton & { radioButtons: typeof data }> = (args) => ({
+  components: { CustomRadioButton },
+  setup() {
+    return { args, options: args.radioButtons };
+  },
+  template: `
+  <form>
+  <FormFieldset>
+  <FormFieldsetLegend>Kies locatie</FormFieldsetLegend>
+  <FormField v-for="{id, label, name} in options" :key="id">
```
Robbert, line 21: I think `name` should not be in this loop, because `name` needs to be the same for all custom radio buttons in the same group. Destructure `value` instead:

`v-for="{id, label, value} in options"`

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonWithFieldsetAndLegend.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <FormLabel :for="id" type="radio" :checked="args.checked" :disabled="args.disabled">
+        <CustomRadioButton :name="name" v-bind="args" :id="id" :value="id" @update:nodeValue="action"/>
```
Robbert, line 23: `:value="id"` → `:value="value"`

Robbert, line 23: Is `nodeValue` correct here, or should it be `modelValue`?

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButton.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+});
+
+export const Default = Template.bind({});
+Default.parameters = {
+  docs: {
+    source: {
+      code: `<CustomRadioButton />`,
+    },
+  },
+};
```
Robbert, line 27: The file name extension for these stories is `.tsx`, but shouldn't it be `.ts`?

AliKdhim87: Honestly, it should be `.vue`, but it's unfortunate I could not find an explanation about Vue.js and Typescript, I don't know if there is a difference between using `.ts` vs `.tsx` with Vuejs,
Should I change them to `.ts` What do you think?

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonWithFieldsetAndLegend.stories.tsx
AliKdhim87: `modelValue` the correct one I will update it

AliKdhim87: Do you mean should I  define the `value` field in the array of object `argTypes` and use it?
I used the `id` as a value  it's just to  print the value 1 or 2 when you toggle between the radio buttons, but if there is a better example I can change it to, just let me know

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonWithFieldsetAndLegend.stories.tsx
AliKdhim87: make sense

packages/storybook-vue/src/stories/CustomRadioButton/CustomRadioButtonStates.stories.tsx
AliKdhim87: I will remove them

Robbert (approved):

PR #1149, opened 2022-08-09
docs: add react backdrop component stories

AliKdhim87:


packages/storybook-react/src/stories/Button/ButtonAppearances.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+export default {
+  title: 'React.js Component/Button/Appearances',
+  id: 'react-button-appearances',
+  component: Button,
+  argTypes,
+} as ComponentMeta<typeof Button>;
+
+const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />;
+
+export const Primary = Template.bind({});
```
Robbert, line 15: Besides `primary-action-button`, `secondary-action-button` and `subtle-button` there is also the default without `appearance` that should be included.

packages/storybook-react/src/stories/Button/argTypes.ts
```diff
@@ -0,0 +1,10 @@
+export const argTypes = {
+  appearance: {
+    type: 'select',
+    options: ['primary-action-button', 'secondary-action-button', 'subtle-button'],
```
Robbert: `undefined` is a missing option here

packages/storybook-react/src/stories/Button/argTypes.ts
```diff
@@ -0,0 +1,10 @@
...
+  },
+  type: {
+    type: 'select',
+    options: ['button', 'submit', 'reset'],
```
Robbert: `undefined` should be an option here too

packages/storybook-react/src/stories/ButtonLink/Button.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import { ComponentMeta, ComponentStory } from '@storybook/react';
+import { ButtonLink } from '@utrecht/component-library-react';
+import React from 'react';
+import { argTypes } from './argTypes';
+
+export default {
+  title: 'React.js Component/ButtonLink',
```
Robbert: In the existing Storybook this is called `Link that looks like a button`

packages/storybook-react/src/stories/ButtonLink/argTypes.ts
```diff
@@ -0,0 +1,10 @@
+export const argTypes = {
+  appearance: {
+    type: 'select',
+    options: ['primary-action-button', 'secondary-action-button', 'subtle-button'],
+  },
+  type: {
+    type: 'select',
+    options: ['button', 'submit', 'reset'],
+  },
```
Robbert, line 13: The `ButtonLink` component doesn't have a `type` argument (since it is not actually a `<button>`)

And the `external` property is missing from `argTypes` (since it is a link)

packages/storybook-react/src/stories/ButtonLink/argTypes.ts
```diff
@@ -0,0 +1,10 @@
+export const argTypes = {
+  appearance: {
+    type: 'select',
+    options: ['primary-action-button', 'secondary-action-button', 'subtle-button'],
```
Robbert: Both the CSS module and the React module for ButtonLink don't yet support `appearance`. We should spend some time in the future to implement this, but until that time we shouldn't have it in Storybook.

packages/storybook-react/src/stories/ButtonLink/Button.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+  argTypes,
+} as ComponentMeta<typeof ButtonLink>;
+
+const Template: ComponentStory<typeof ButtonLink> = (args) => <ButtonLink {...args} />;
+
+export const Default = Template.bind({});
+
+Default.args = {
+  children: 'Read more...',
+};
```
Robbert: The files in `ButtonLink/` are called `Button*.tsx` instead of `ButtonLink*.tsx`

Robbert (approved):

PR #1161, opened 2022-08-11
docs: add react html content component stories

AliKdhim87:


Robbert:
@AliKdhim87 I pushed some tweaks, could you review those tomorrow?

packages/storybook-react/src/stories/OrderedList/argTypes.ts
```diff
@@ -0,0 +1,26 @@
+export const argTypes = {
+  reversed: {
+    type: { name: 'boolean', required: false },
+    table: {
+      category: 'HTML attribute',
```
AliKdhim87, line 5: This is really nice!

packages/storybook-react/src/stories/OrderedList/OrderedListItem.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+} as ComponentMeta<typeof OrderedListItem>;
+
+export const Default: ComponentStory<typeof OrderedListItem> = (args) => <OrderedListItem {...args} />;
+
+Default.args = {
+  children: 'List item',
+};
+
+// Configure a `displayName` for the code snippet when used as Storybook subcomponent
+Default.displayName = 'OrderedListItem';
```
AliKdhim87, line 21: TypeScript complains about this locally 
```ts
Property 'displayName' does not exist on type 'ComponentStory<ForwardRefExoticComponent<OrderedListItemProps & RefAttributes<HTMLLIElement>>>'.ts(2339)
```

Also, do you think we need it?
Because we already set the  `displayName` in the [React component implementation](https://github.com/nl-design-system/utrecht/blob/main/packages/component-library-react/src/OrderedListItem.tsx#L22)

Robbert: The `displayName` is for the Storybook Template component that wraps our component that indeed already has the display name.

I added it to fix this issue: https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097

PR #1166, opened 2022-08-12
docs: add react paragraph component stories

AliKdhim87:


packages/storybook-react/src/stories/RadioButton/RadioButtonStates.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const Focus = Template.bind({});
+
+Focus.parameters = {
+  pseudo: { focus: true },
+};
+
+export const FocusVisible = Template.bind({});
+
+FocusVisible.parameters = {
+  pseudo: { focusVisible: true },
```
Robbert, line 30: I think for a realistic version we should always use `focus` together with `focusVisible`:

```js
{
  pseudo: {
    focus: false,
    focusVisible: true,
  },
}
```

We should probably also look back at previous code to fix this for other pseudo states.

Robbert (approved):

PR #1178, opened 2022-08-15
feat: add html-select css class

AliKdhim87:


packages/component-library-react/src/Select.test.tsx
```diff
@@ -175,22 +158,14 @@ describe('Select', () => {
     });
 
     it('omits non-essential required attributes when not required', () => {
-      const { container } = render(<Select required={false} />);
```
Robbert, line 178: `required={false}` is needed here, we specifically want test that for attributes that are false by default we don't clutter the HTML.

packages/storybook-react/src/stories/Surface/Separator.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+  component: Surface,
+} as ComponentMeta<typeof Surface>;
+
+const Template: ComponentStory<typeof Surface> = (args) => <Surface {...args} />;
+
+export const Default = Template.bind({});
+
+Default.args = {
+  children: 'The Quick Brown Fox Jumps Over The Lazy Dog',
+};
```
Robbert, line 17: `Surface/Separator.stories.tsx` ➝ `Surface/Surface.stories.tsx`

packages/storybook-react/src/stories/Select/argTypes.ts
```diff
@@ -0,0 +1,49 @@
...
+    type: { name: 'string' },
+    table: {
+      category: 'HTML attribute',
+      defaultValue: { summary: '' },
+    },
+  },
+};
+
+export const options = [
+  { value: '', label: 'Select an option', disabled: true },
```
Robbert, line 44: Maybe we should investigate the placeholder option some more, for a separate PR.

Would be nice to be able to do `<Select placeholder="Select an option">`. The placeholder option should ideally use a `placeholder-color` design token instead of `disabled-color`.

So if the placeholder attribute could render the following it could be beneficial:

```html
<select class="utrecht-select utrecht-select--placeholder-selected">
  <option value="" disabled selected class="utrecht-select__option utrecht-select__option--placeholder">Placeholder text</option>
</select>
```

packages/storybook-react/src/stories/Select/SelectStates.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+Disabled.args = {
+  disabled: true,
+};
+
+export const Required = Template.bind({});
+
+Required.args = {
+  required: true,
+  value: '',
```
Robbert, line 35: Can `value: '',` be removed? It clutters the code snippet.

packages/storybook-react/src/stories/Select/SelectStates.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const RequiredServerSideRendering = Template.bind({});
+
+RequiredServerSideRendering.args = {
+  required: true,
+  noscript: true,
+  value: '',
```
Robbert, line 43: Can `value: '',` be removed?

Robbert (changes requested):

Robbert (approved):

PR #1179, opened 2022-08-15
ci: replace skip percy

Yolijn:


components/icon/web-component/stories/default.stories.mdx
```diff
@@ -39,10 +39,6 @@ import { IconExample } from "../template";
       icon: "utrecht-icon-loupe",
       size: 32,
     }}
-    parameters={{
-      // This is an interactive example where styling is unimportant, ignore this test
-      percy: false,
```
Yolijn, line 44: I'm not sure if this should now be `chromatic: { disableSnapshot: true },` or not

Yolijn: @Robbert was this a different way of disabling percy tests for this component?

Robbert (approved):

PR #1199, opened 2022-08-22
ci: use multi semantic release instead of lerna

AliKdhim87:


Robbert:
Closing this for now, need to reconsider approach now that we use `pnpm`.

PR #1201, opened 2022-08-23
feat: improve the custom checkbox component css

AliKdhim87:


components/custom-checkbox/css/_mixin.scss
```diff
@@ -3,107 +3,123 @@
...
-}
-
-@mixin utrecht-custom-checkbox__box--invalid {
+@mixin utrecht-custom-checkbox--indeterminate {
   background-color: var(
-    --utrecht-custom-checkbox-invalid-background-color,
+    --utrecht-custom-checkbox-indeterminate-background-color,
     var(--utrecht-custom-checkbox-background-color)
   );
+  background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'%3e%3cpath fill='none' stroke='%23fff' stroke-linecap='round' stroke-linejoin='round' stroke-width='3' d='M6 10h8'/%3e%3c/svg%3e");
```
dutchcelt, line 58: All the mixins (and mixins in general) use variables in some way and don't contain hard-coded asset references or property values. Except for these data URLs. 

Is it possible to keep the SVG image in the markup instead?  
Using a single icon component for the multiple states. 
The other option is to move the data value to the proprietary tokens. That way, the icons remain themable.

Robbert: I think this approach with the hardcoded `<svg>` was meant to achieve the desired styling with only `<input type="checkbox">`. For the Web Component/React etc variant having the SVG in the markup is absolutely fine. For web components I guess it would be nice to able to customize the icon with `<svg slot="checkmark-checked"/>` or something like that.

Robbert (approved):

PR #1204, opened 2022-08-23
docs: fix previews source

Yolijn:
- Use transformSrc to ensure valid HTML instead of react templates in the preview source of stories
- To ensure use of the current parameters of a story in tranformSource: pass them on in the preview.js decorator (hopefully this can be removed later once Storybook solves the issue themselves)
- Use `ReactDOMServer.renderToStaticMarkup(element)` to get the valid HTML output of the story component. For this to work we need the component available in the story context

Todo: add components for all stories while transforming them into React components


components/button/css/stories/default.stories.mdx
```diff
@@ -1,5 +1,6 @@
 <!-- @license CC0-1.0 -->
 
+import * as ReactDOMServer from "react-dom/server";
```
Yolijn, line 3: Oops, this can be removed

Robbert (approved):

PR #1246, opened 2022-09-09
style: allowing for shared properties for data lists

dutchcelt:
Addition to the data-list styling that enables the Den Haag definition-list to be emulated without breaking the existing styling of the data-list.

### DIFF
--- a/components/data-list/css/_mixin.scss
+++ b/components/data-list/css/_mixin.scss
@@ -19,7 +19,6 @@
   color: var(--utrecht-data-list-item-key-color);
   font-size: var(--utrecht-data-list-item-key-font-size);
   font-weight: var(--utrecht-data-list-item-key-font-weight);
-  grid-area: label;
   line-height: var(--utrecht-data-list-item-key-line-height);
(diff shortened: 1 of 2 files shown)


components/data-list/css/_mixin.scss
```diff
@@ -19,18 +19,17 @@
...
   font-size: var(--utrecht-data-list-item-key-font-size);
   font-weight: var(--utrecht-data-list-item-key-font-weight);
-  grid-area: label;
   line-height: var(--utrecht-data-list-item-key-line-height);
 }
 
 @mixin utrecht-data-list-item-value {
   /* configure `min-block-size` in case the value is empty */
-  color: var(--utrecht-data-list-label-color);
+  color: var(--utrecht-data-list-item-value-color, var(--utrecht-data-list-label-color));
```
Robbert, line 27: We can simply remove the `--utrecht-data-list-label-color`, the `label` was a bug.

Robbert (changes requested):

PR #1247, opened 2022-09-10
refactor: move web components code from components/ to own package

Robbert:


components/icon/package.json
```diff
@@ -11,9 +11,9 @@
   "scripts": {
```
AliKdhim87, line 11: I think we need to remove `loader/` from `files: []`
https://github.com/nl-design-system/utrecht/pull/1247/files#diff-c348480d87c6f739dfc4ad2b62910527ce30b8ee41c93897b2f8b37a8d6e5861R9

Robbert: Why?

AliKdhim87: Because we don't have `loader ` folder as output in the package

PR #1252, opened 2022-09-14
feat(link-social-react): create link social component

AliKdhim87:


components/link-social/css/index.scss
```diff
@@ -5,21 +5,24 @@
...
   --utrecht-icon-size: var(--utrecht-link-social-icon-size);
+  --utrecht-icon-color: var(--utrecht-link-social-color);
 
   align-items: center;
   background-color: var(--utrecht-link-social-background-color);
   border-color: var(--utrecht-link-social-border-color);
   border-radius: 50%;
   border-style: solid;
   border-width: var(--utrecht-link-social-border-width);
-  color: var(--utrecht-link-social-color);
```
Robbert, line 15: I would rather leave the `color` so a text a hypothetical text alternative to the icon automatically has enough contrast.

components/link-social/css/stories/default.stories.mdx
```diff
@@ -6,7 +6,7 @@ import "../index.scss";
 
 <Meta
   id="css-link-social"
-  title="CSS Component/Link to Social Media"
+  title="CSS Component/Link Social"
```
Robbert, line 9: Isn't "Link to Social Media" more helpful to readers of the documentation than "Link Social"?

I would prefer to keep the long title.

We might even put the `list-social` under `"Link to Social Media/Link list"`

PR #1283, opened 2022-09-26
style: allow to use breadcrumb component overrides

dutchcelt:
The changes here are to allow RVO use the Utrecht breadcrumbs styles and tokens without breaking either implementation.

### DIFF
--- a/components/breadcrumb/css/_mixin.scss
+++ b/components/breadcrumb/css/_mixin.scss
@@ -21,6 +21,7 @@
   --utrecht-link-visited-color: var(--utrecht-breadcrumb-link-color);
   --utrecht-link-visited-text-decoration: var(--utrecht-link-text-decoration);
 
+  align-items: baseline;
   font-family: var(--utrecht-document-font-family, inherit);
   font-size: var(--utrecht-breadcrumb-font-size);
   text-transform: var(--utrecht-document-text-transform, inherit);
@@ -32,7 +33,12 @@
 }
 
 @mixin utrecht-breadcrumb__item {
+  align-items: baseline;
   block-size: 100%;
+  display: inline-flex;
+  & > * {
+    margin-inline-end: var(--utrecht-breadcrumb-item-gap, 0);
+  }
 }
 
 @mixin utrecht-breadcrumb__link {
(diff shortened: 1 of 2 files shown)


components/breadcrumb/css/_mixin.scss
```diff
@@ -42,70 +48,20 @@
...
-
-  overflow: hidden;
-
-  .utrecht-breadcrumb__link {
-    padding-inline-end: 0;
-    position: relative;
+  text-decoration: var(--utrecht-breadcrumb-link-text-decoration);
+  &:hover {
+    text-decoration: var(--utrecht-breadcrumb-link-hover-text-decoration);
+    text-decoration-thickness: var(--utrecht-breadcrumb-link-hover-text-decoration-thickness);
```
Robbert, line 54: This change breaks the existing hover style of Utrecht, the thick underline.

In addition to the CSS changes, it would be great if you would include stories for that simulate these hover states, so we have visual regression tests to prevent this situation in the future.

Robbert, line 54: It would be more ideal if you would implement these breadcrumb specific link tokens similar to how the existing code works: 

```css
--utrecht-link-hover-text-decoration: var(--utrecht-breadcrumb-link-hover-text-decoration);
```

If you don't specify the `breadcrumb-link` tokens, I would want the links to render like the default link component.

proprietary/design-tokens/src/component/utrecht/breadcrumb.tokens.json
```diff
@@ -27,19 +30,42 @@
...
+          "text-decoration": {
+            "value": "underline"
+          },
+          "text-decoration-thickness": {
+            "value": "1px"
           }
+        },
+        "text-decoration": {
+          "value": "underline"
         }
```
Robbert, line 69: I'm not exactly sure why this change for the benefit of RVO would require token changes for Utrecht. If it is necessary to maintain existing behavior, that would be a good indication this is a breaking change.

components/breadcrumb/css/_mixin.scss
```diff
@@ -32,7 +33,12 @@
 }
 
 @mixin utrecht-breadcrumb__item {
+  align-items: baseline;
   block-size: 100%;
+  display: inline-flex;
+  & > * {
+    margin-inline-end: var(--utrecht-breadcrumb-item-gap, 0);
+  }
```
Robbert, line 41: If this is to remove a margin on link, I would rather have the property on `__link`.

Robbert (changes requested):

PR #1284, opened 2022-09-27
style: extending the usage to rvo design system

dutchcelt:


components/data-list/css/_mixin.scss
```diff
@@ -16,10 +20,17 @@
 }
 
 @mixin utrecht-data-list-item-key {
+  border-right-color: var(--utrecht-data-list-item-key-border-right-color, transparent);
+  border-right-style: solid;
+  border-right-width: var(--utrecht-data-list-item-key-border-right-width, 0);
```
Robbert, line 25: Let's make these logical border properties.

proprietary/design-tokens/src/component/utrecht/data-list.tokens.json
```diff
@@ -1,26 +1,40 @@
...
-        "line-height": {}
+        "line-height": {},
+        "padding-block-end": {},
+        "padding-block-start": {},
+        "padding-inline-end": {},
+        "padding-inline-start": {}
       },
       "rows": {
+        "border-bottom-color": {},
+        "border-bottom-width": {},
```
Robbert, line 37: All these empty `{}` don't have much use here for the Utrecht theme, but you might have included them so they contribute to the automatically generated design tokens documentation.

All new tokens should be included in the `tokens.json` file next in the `components/` package, in this case here:
https://github.com/nl-design-system/utrecht/blob/main/components/data-list/tokens.json

components/data-list/css/_mixin.scss
```diff
@@ -29,6 +40,10 @@
   font-weight: var(--utrecht-data-list-item-value-font-weight);
   line-height: var(--utrecht-data-list-item-value-line-height);
   min-block-size: calc(var(--utrecht-data-list-item-value-line-height) * 1rem);
+  padding-block-end: var(--utrecht-data-list-item-value-padding-block-end, 0);
+  padding-block-start: var(--utrecht-data-list-item-value-padding-block-start, 0);
+  padding-inline-end: var(--utrecht-data-list-item-value-padding-inline-end, 0);
+  padding-inline-start: var(--utrecht-data-list-item-value-padding-inline-start, 0);
```
Robbert, line 46: I suspect the padding should also be configurable on the actions. What do you think?

dutchcelt: Yes, i've added these to actions as well. This does imply we need another vertical border. Haven't added it as it doesn't affect the Utrecht implementation. We may need to refactor this implementation after the PR to clean it up.

PR #1401, opened 2022-11-29
feat: create accordion css component + stories

AliKdhim87:


packages/component-library-react/src/Accordion/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+          appearance="subtle-button"
+          aria-expanded={expanded}
+          aria-controls={`utrecht-accordion-label-${id}`}
+          id={`${id}-id`}
+          onClick={() => setExpanded(!expanded)}
+        >
+          {label}
+        </Button>
+      </Heading3>
+      {expanded && (
```
Robbert, line 31: When you only render the expanded sections, search engines will not be able to index collapsed accordion sections.

PR #1489, opened 2023-01-10
test: add test to the angular heading component

AliKdhim87:


Robbert (commented):
The heading components don't render correctly in Storybook. The `utrecht-heading-1` attribute is rendered as-is, instead of making use of the Angular component and adding the `class="utrecht-heading-1"`. Same for the other headings.

Screenshot 2023-01-11 at 09 40 38

Robbert:
I think this component should probably be rendered as `<utrecht-heading-1>` instead of `<h1 utrecht-heading-1>`.

Robbert (approved):

Robbert (approved):

PR #1492, opened 2023-01-12
test: add test to the checkbox angular component

AliKdhim87:


packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    expect(fixture.nativeElement).toBeVisible();
+    expect(fixture.nativeElement).toHaveStyle({ display: 'block' });
+  });
+
+  it('renders a design system BEM block class name', async () => {
+    const { fixture } = await render('<input type="checkbox" utrecht-checkbox />', {
+      declarations: [UtrechtCheckboxAttr],
+    });
+    const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input'))!;
```
Robbert, line 43: Isn't the other way around the convention? `input[type="checkbox"]`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+    expect(debugElement.nativeElement.classList.contains('utrecht-checkbox')).toBeTruthy();
+  });
+
+  describe('checked variant', () => {
+    it('is not checked by default', async () => {
+      const { fixture } = await render('<input type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+      });
+
+      expect(fixture.nativeElement).not.toBeChecked();
```
Robbert, line 53: For accessibility tree based checks, such as `toBeChecked()`, I would prefer you locate the relevant element using `getByRole('checkbox')`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    it('omits non-essential checked attributes when not checked', async () => {
+      const handleChange = jest.fn();
+      const { fixture } = await render(
+        '<input (ngModelChange)="handleChange()" checked="false" type="checkbox" utrecht-checkbox />',
+        {
+          declarations: [UtrechtCheckboxAttr],
+        },
+      );
+      const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input'));
```
Robbert, line 64: I also think `getByRole()` would be nicer here too.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+      const handleChange = jest.fn();
+      const { fixture } = await render(
+        '<input (change)="handleChange()" checked="true" type="checkbox" utrecht-checkbox />',
+        {
+          declarations: [UtrechtCheckboxAttr],
+        },
+      );
+      fixture.detectChanges();
+      const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input'));
+      handleChange();
```
Robbert, line 81: I don't think your testing any part of the component by calling `handleChange` yourself. How about using a testing library function to click the radio button?

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+      const handleChange = jest.fn();
+      const { fixture } = await render(
+        '<input (change)="handleChange()" checked="true" type="checkbox" utrecht-checkbox />',
+        {
+          declarations: [UtrechtCheckboxAttr],
+        },
+      );
+      fixture.detectChanges();
+      const debugElement = fixture.debugElement.query(By.css(':checked'));
+      handleChange();
```
Robbert, line 97: For the `:checked` test, I wouldn't expect we need to have any event listeners. When it is checked after rendering, that should be enough.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+      expect(handleChange).toHaveBeenCalled();
+
+      expect(debugElement.nativeElement).toBeInTheDocument();
+    });
+  });
+  describe('invalid variant', () => {
+    it('renders a design system BEM modifier class name', async () => {
+      const { fixture } = await render('<input invalid="true" type="checkbox" utrecht-checkbox />', {
```
Robbert, line 106: In Angular, for rendering boolean properties, I remember you need to use `[invalid]="true"` instead of `invalid="true"`. The first will be a boolean, because the contents of the attributes are evaluated as JavaScript. The second would be a `"true"` string.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+    });
+  });
+
+  describe('disabled variant', () => {
+    it('renders a design system BEM modifier class name', async () => {
+      const { fixture } = await render('<input disabled="true" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+      });
+      fixture.detectChanges();
+      const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input:disabled'));
```
Robbert, line 163: `input[type="checkbox"]:disabled` is more conventional I think.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+      expect(debugElement.nativeElement).toHaveClass('utrecht-checkbox--disabled');
+    });
+
+    it('is not disabled by default', async () => {
+      const { fixture } = await render('<input type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+      });
+      fixture.detectChanges();
+      const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input'));
```
Robbert, line 173: `getByRole()` would be more appropriate than `query`.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+      const debugElement = fixture.debugElement.query(By.css('[type="checkbox"]input'));
+
+      fireEvent.input(debugElement.nativeElement, {
+        target: {
+          checked: 'true',
+          value: 'checked',
+        },
+      });
+      fixture.detectChanges();
```
Robbert, line 338: Again, would it be possible to trigger this indirectly, by simulating a `click` or `Enter` through testing library?

packages/storybook-angular/src/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+      },
+    },
+    disabled: {
+      name: 'disabled',
+      type: { name: 'boolean', required: false },
+      table: {
+        category: 'HTML',
+        defaultValue: { summary: false },
+      },
+    },
```
Robbert, line 24: I think we should include `invalid` and `required` in argTypes too.

packages/storybook-angular/src/stories/CheckboxStates.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+Hover.parameters = {
+  pseudo: { hover: true },
+};
+
+export const Focus = Template.bind({});
+
+Focus.parameters = {
+  pseudo: { focus: true, focusVisible: true },
+};
```
Robbert, line 47: I suggest we put all the checkbox stories in one file, `Checkbox.stories.tsx`.

I think I mainly want to use separate files for stories that combine multiple components, instead of showing of the atomic component.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
AliKdhim87: Yes, of course!
I'm gonna update them!

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    const checkbox = screen.getByRole('checkbox');
+
+    expect(checkbox).toBeInTheDocument();
+    expect(checkbox).toBeVisible();
+  });
+
+  it('renders an HTML input type=checkbox element', async () => {
+    const { fixture } = await render('<input type="checkbox" utrecht-checkbox />', {
+      declarations: [UtrechtCheckboxAttr],
+    });
```
Robbert, line 25: There should at least be `By.query('input[type="checkbox"]')` to run the test.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    it('is not checked by default', async () => {
+      await render('<input type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+      });
+      const checkbox = screen.getByRole('checkbox');
+      expect(checkbox).not.toBeChecked();
+    });
+
+    it('omits non-essential checked attributes when not checked', async () => {
+      await render('<input checked="false" type="checkbox" utrecht-checkbox />', {
```
Robbert, line 56: This should probably be `[checked]="false"` and `not.toHaveAttribute('checked')`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    it('can have a checked state', async () => {
+      const handleChange = jest.fn();
+      await render('<input (change)="sendValue($event)" [checked]="checked" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          sendValue: handleChange,
+          checked: true,
+        },
+      });
```
Robbert, line 73: I think the approach for a checked state is great, but the `(change)` event handler should be moved to a completely separate test. Has nothing to do with the checked state per se.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+      const checkbox = screen.getByRole('checkbox');
+      checkbox?.click();
+
+      expect(handleChange).toHaveBeenCalledTimes(1);
+      expect(checkbox).toBeChecked();
+    });
+
+    it('can have a checked state in CSS', async () => {
+      const { container } = await render('<input checked="true" type="checkbox" utrecht-checkbox />', {
```
Robbert, line 83: `[checked]="true"`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+  describe('invalid variant', () => {
+    it('renders a design system BEM modifier class name', async () => {
+      const { container } = await render('<input [invalid]="invalid" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          invalid: true,
+        },
+      });
+
+      const checkbox = container.querySelector('input[type="checkbox"]');
```
Robbert, line 100: I suppose it would be more to the point to use `querySelector('.utrecht-checkbox')` in these BEM modifier class name tests.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+  describe('disabled variant', () => {
+    it('renders a design system BEM modifier class name', async () => {
+      const { container } = await render('<input [disabled]="disabled" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          disabled: true,
+        },
+      });
+
+      const checkbox = container.querySelector('input[type="checkbox"]:disabled');
```
Robbert, line 162: By using `:disabled` here, you are testing two things at once. I suggest to simply use `querySelector('.utrecht-checkbox')` here.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    });
+
+    it('can have a disabled state in CSS', async () => {
+      const { container } = await render('<input [disabled]="disabled" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          disabled: true,
+        },
+      });
+      const checkbox = container.querySelector('input[type="checkbox"]:disabled');
```
Robbert, line 209: I think that we can keep this selector simple, like you did in the other tests: `querySelector(':disabled')`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+  describe('required variant', () => {
+    it('renders a design system BEM modifier class name', async () => {
+      const { container } = await render('<input [required]="required" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          required: true,
+        },
+      });
+      const checkbox = container.querySelector('input[type="checkbox"]:required');
```
Robbert, line 223: `'input[type="checkbox"]:required'` ➝ `'.utrecht-checkbox'`?

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    it('can have a required state', async () => {
+      const { container } = await render('<input [required]="required" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          required: true,
+        },
+      });
+
+      const checkbox = container.querySelector('input[type="checkbox"]:required');
```
Robbert, line 258: `querySelector('input[type="checkbox"]:required')` ➝  `getByRole('checkbox')`?

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    it('can have a required state in CSS', async () => {
+      const { container } = await render('<input [required]="required" type="checkbox" utrecht-checkbox />', {
+        declarations: [UtrechtCheckboxAttr],
+        componentProperties: {
+          required: true,
+        },
+      });
+
+      const checkbox = container.querySelector('input[type="checkbox"]:required');
```
Robbert, line 271: Simply `:required` would suffice.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+      expect(checkbox).toBeInTheDocument();
+    });
+  });
+
+  it('can be hidden', async () => {
+    const { container } = await render('<input [hidden]="[hidden]" type="checkbox" utrecht-checkbox />', {
```
Robbert, line 278: `[hidden]="[hidden]"` ➝ `[hidden]="hidden"

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    const { container } = await render('<input [class]="[class]" type="checkbox" utrecht-checkbox />', {
+      declarations: [UtrechtCheckboxAttr],
+      componentProperties: {
+        class: 'custom-class',
+      },
+    });
+
+    const checkbox = container.querySelector('input[type="checkbox"]');
+
+    expect(checkbox).toHaveClass('custom-class');
```
Robbert, line 300: Please update this to do a slightly different test: `'can have an additional class name'`. The also check that the BEM class name is still there: the component must combine class names, not overwrite the BEM class names.

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+  });
+
+  describe('change event', () => {
+    it('can trigger a change event', async () => {
+      const handelChange = jest.fn();
+      const { container } = await render(
+        '<input type="checkbox" (change)="handelChange($event)" [checked]="checked" utrecht-checkbox />',
```
Robbert, line 307: `handelChange` ➝ `handleChange`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+            handelChange,
+          },
+        },
+      );
+
+      const checkbox = container.querySelector('input[type="checkbox"]');
+      (checkbox as HTMLInputElement)?.click();
+
+      expect(handelChange).toHaveBeenCalledTimes(0);
+      expect(checkbox).toBeDisabled();
```
Robbert, line 340: `toBeDisabled()` ➝ `not.toBeChecked()`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    });
+  });
+
+  describe('input event', () => {
+    it('can trigger a input event', async () => {
+      const handelInput = jest.fn();
+
+      const { container } = await render(
+        '<input (input)="handelInput($event)" type="checkbox" [checked]="checked" utrecht-checkbox />',
```
Robbert, line 349: - `handelInput` ➝ `handleInput`
- `[checked]="checked"` can be omitted

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
AliKdhim87: I'm gonna use  `container.querySelector('input[type="checkbox"]')`

packages/component-library-angular/src/checkbox/Checkbox.spec.ts
AliKdhim87: I'm gonna update the test text here, also I will add `expect(checkbox).toHaveClass('utrecht-checkbox--html-input utrecht-checkbox--custom');` addition check to make sure we still have the another classes, and here is the output from the `debug()` function

```
    <input
      class="custom-class utrecht-checkbox--html-input utrecht-checkbox--custom utrecht-checkbox"
      type="checkbox"
      utrecht-checkbox=""
    />
```html

Robbert (approved):
I think you did a beautiful job here, thank you!!

PR #1493, opened 2023-01-12
test: add test to the document angular component

AliKdhim87:


packages/component-library-angular/src/document/Document.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(doc).toHaveStyle({ display: 'block' });
+  });
+
+  it('renders rich text content', async () => {
+    const fixture = TestBed.createComponent(UtrechtDocument);
+    const debugElement = fixture.debugElement.query(By.css('div'))!;
+    debugElement.nativeElement.innerHTML = '<h1>Breaking news</h1>';
+    fixture.detectChanges();
+    expect(debugElement.nativeElement).toContainHTML('h1');
+  });
```
Robbert, line 43: I don't think this approach for rendering rich text is a realistic test, we need to render it through Angular APIs. This test would still succeed even when we would remove `<ng-content/>`

packages/component-library-angular/src/document/component.ts
```diff
@@ -7,5 +7,6 @@ import { ChangeDetectionStrategy, Component } from '@angular/core';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class UtrechtDocument {
+  @Input() class?: string;
```
Robbert, line 10: I want to avoid adding global HTML attributes such `class` to our Angular components, so I would propose to remove this `@Input()`. If someone really needs `utrecht-document` with custom class names, then we should probably offer `<div utrecht-document>` instead.

packages/component-library-angular/src/document/Document.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    const { container } = await render('<utrecht-document hidden="true"></utrecht-document>', {
+      declarations: [UtrechtDocument],
+    });
+
+    const doc = container.querySelector(':only-child');
+
+    expect(doc).not.toBeVisible();
+  });
+
+  it('can have a custom class name', async () => {
```
Robbert, line 55: I think we should remove this test for now.

packages/component-library-angular/src/document/component.ts
AliKdhim87: Thanks for this feedback, I will remove it
I found how to add a modifier class in Angular without using the `@Input() class`, actually use 
`<div utrecht-document [class]="custom-class">`

PR #1543, opened 2023-02-02
feat: add button-link angular component & add test

AliKdhim87:


packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      declarations: [UtrechtButtonLinkAttr],
+    });
+
+    const buttonLink = screen.getByRole('link', { name: 'Link' });
+
+    expect(buttonLink).toBeInTheDocument();
+    expect(buttonLink).toBeVisible();
+  });
+
+  it('renders an HTML button link element', async () => {
```
Robbert, line 17: `utrecht-button` ➝ `utrecht-button-link`

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+        hidden: true,
+      },
+    });
+
+    const buttonLink = container.querySelector(':only-child');
+
+    expect(buttonLink).not.toBeVisible();
+  });
+
+  it('can have a addition class name', async () => {
```
Robbert, line 67: `a addition` ➝ `an additional`

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+    const { container } = await render('<a href="https://example.com" [class]="class" utrecht-button-link></a>', {
```
Robbert, line 68: Is there a reason to use `[class]="class"` instead of simply `class="custom-class"`? The code would probably be a bit more straightforward to understand. Same goes for `hidden`, I suppose.

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      componentProperties: {
+        class: 'custom-class',
+      },
+    });
+
+    const buttonLink = container.querySelector(':only-child');
+
+    expect(buttonLink).toHaveClass('custom-class');
+    expect(buttonLink).toHaveClass('utrecht-button-link');
+    expect(buttonLink).toHaveClass('utrecht-button-link--html-a');
```
Robbert, line 79: I don't think you need to test for `utrecht-button-link--html-a`, since I don't consider it to be a public API. We should be able to refactor the CSS selector from `utrecht-button-link--html-a` to `a.utrecht-button-link` without anyones code breaking. But we can never change `utrecht-button-link` that way.

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  describe('Button Link Appearance', () => {
+    it('renders a design system BEM `utrecht-button-link--primary-action` class', async () => {
+      const { container } = await render(
+        '<a href="https://example.com" [appearance]="appearance" utrecht-button-link></a>',
```
Robbert, line 85: Same here: could we simply have `appearance="primary-action-button"`?

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+          componentProperties: {
+            appearance: 'subtle-button',
+          },
+        },
+      );
+      const buttonLink = container.querySelector('.utrecht-button-link');
+
+      expect(buttonLink).toHaveClass('utrecht-button-link--subtle-action');
+    });
+    it('can not renders a design system BEM `utrecht-button-link--primary-action` class by default', async () => {
```
Robbert, line 126: `can not renders` ➝ `does not have`

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      const { container } = await render('<a href="https://example.com" utrecht-button-link></a>', {
+        declarations: [UtrechtButtonLinkAttr],
+      });
+      const buttonLink = container.querySelector('.utrecht-button-link');
+
+      expect(buttonLink).not.toHaveClass('utrecht-button-link--subtle-action');
+    });
+  });
+
+  describe('Button Link `external`', () => {
```
Robbert, line 153: You don't have to repeat `Button Link` in nested `describe`s.

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      await render('<a utrecht-button-link [external]="external" href="https://example.com">Link</a>', {
+        declarations: [UtrechtButtonLinkAttr],
+        componentProperties: {
+          external: true,
+        },
+      });
+
+      const buttonLink = screen.getByRole('link', { name: 'Link' });
+
+      expect(buttonLink).toHaveAttribute('rel', 'external noopener noreferer');
```
Robbert, line 164: The order of the relationship tokens are not relevant, so I would try to test for this in another way. For example using the attribute value selector that works specifically for whitespace separated lists: `a[rel~="external" i]`.

For more specific error messages, I would make a separate `expect()` for every value.

The `prevents sharing referer info` should only check for `noreferer` by the way.

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    it('prevents access to window.opener', async () => {
+      await render('<a utrecht-button-link [external]="external" href="https://example.com">Link</a>', {
+        declarations: [UtrechtButtonLinkAttr],
+        componentProperties: {
+          external: true,
+        },
+      });
+      const buttonLink = screen.getByRole('link', { name: 'Link' });
+      expect(buttonLink.getAttribute('rel')).toContain('noopener');
```
Robbert, line 175: `toContain('noopener')` would also work if the attribute was `rel="brokennoopener"`, so I would use a method that actually works with values in a whitespace separated list.

packages/storybook-angular/src/stories/ButtonLink.stories.ts
```diff
@@ -0,0 +1,183 @@
...
+};
+
+export default buttonLinkMeta;
+type Story = StoryObj<ButtonLinkType>;
+
+const Template: Story = {
+  render: (args) => ({
+    props: args,
+    template:
+      '<a [external]="external" [appearance]="appearance" href="https://example.com/" utrecht-button-link>Read More</a>',
```
Robbert, line 57: Should `href` be `[href]="href"`?

Robbert (changes requested):

packages/component-library-angular/src/button-link/ButtonLink.spec.ts
AliKdhim87: Actually, it works fine without `[class]="class"` also easier, the reason why I used it, is I thought it was like when we export a custom tag like `<utrecht-button-link />`

packages/storybook-angular/src/stories/ButtonLink.stories.ts
AliKdhim87: Of course

Robbert (approved):

PR #1563, opened 2023-02-13
feat: create radio-button-form-field react component & add stories

AliKdhim87:


packages/component-library-react/src/CheckboxFormField.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <FormLabel className={clsx(label.className)} type="checkbox" htmlFor={label.htmlFor} disabled={label.disabled}>
+          {label.text}
+        </FormLabel>
+      )}
+      {description && (
+        <FormFieldDescription className={clsx(description.className)} id={description.id}>
+          {description.message}
+        </FormFieldDescription>
+      )}
+      {validation && validation.message && (
```
sergei-maertens, line 149: It'd be really cool if *multiple* validation errors would be supported - we're working with Form.io which runs multiple validators in parallel which can result in more than one error message being displayed.

Robbert:
Deze feature is gemerged als onderdeel van een andere PR:
https://github.com/nl-design-system/utrecht/tree/main/packages/components-react/radio-group-react

PR #1586, opened 2023-02-21
feat: add angular unit tests and storybook entry for form fieldset component

AndreaBusse29:


packages/component-library-angular/src/form-fieldset/FormFieldset.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    const { container } = await render('<fieldset utrecht-fieldset name="first and last name fields"></fieldset>', {
+      declarations: [UtrechtFieldsetAttr],
+    });
+    const fieldset = container.querySelector('fieldset:only-child');
+    const name = fieldset?.getAttribute('name');
+
+    expect(name).toEqual('first and last name fields');
+  });
+
+  it('can have a form', async () => {
```
Yolijn, line 89: I don't understand this test... what is a form attribute and why are we testing it?

AndreaBusse29: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attributes it's this one, a standard attribute of the fieldset html element.

Yolijn (approved):

PR #1692, opened 2023-04-18
feat: betere HTML voor blockquote

Robbert:
TODO:
- [x] pnpm changeset
- [x] restore existing class names for backwards compatibility, deprecate them
- [x] add fallback for attribution design tokens
- [x] add row-gap token


Robbert:
https://toot.cafe/@aardrian/110770435515706072

PR #1699, opened 2023-04-19
feat: datepicker navigation styling

bddjong:
This change adds styling to the navigation of the datepicker according to the figma design.
image

### Changes
- Add correct colors
- Update font size
- Remove width property and use minimum width required by content of calendar (this should scale much more logically than an arbitrary width number that could conflict with the content size.)
- Add spacing between month label and buttons

[Figma design](https://www.figma.com/file/msb3CfQBefPoruqNQ968Zh/Utrecht-Design-System?node-id=2903-6360&t=VAIWYtcANbzpWIkn-0)

### DIFF
--- a/components/calendar/css/_mixin.scss
+++ b/components/calendar/css/_mixin.scss
@@ -1,6 +1,6 @@
 @mixin utrecht-calendar {
   background-color: var(--utrecht-calendar-background-color);
-  width: var(--utrecht-calendar-width);
+  width: fit-content;
 }
 
 @mixin utrecht-calendar__table {
@@ -92,14 +92,14 @@
 @mixin utrecht-calendar__navigation-label {
   color: var(--utrecht-calendar-navigation-label-color);
   font-size: var(--utrecht-calendar-navigation-label-min-font-size);
-  min-inline-size: var(--utrecht-calendar-navigation-label-min-inline-size);
   text-align: center;
 }
 
 @mixin utrecht-calendar__navigation-buttons {
   align-items: center;
   display: flex;
-  justify-content: center;
+  flex-grow: 1;
+  justify-content: space-between;
 }
 
 @mixin utrecht-calendar__button {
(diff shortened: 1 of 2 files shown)


components/calendar/css/_mixin.scss
```diff
@@ -10,6 +10,9 @@
 }
 
 @mixin utrecht-calendar__table-weeks-item {
+  line-height: var(--utrecht-calendar-table-weeks-item-line-height, inherit);
```
bddjong, line 13: Could effectively do the same without editing the line height. Editing lineheight just makes it easier to stay in line with the design from Figma, and I imagine that also goes for other organisations so this token would be useful. (Den Haag design for example has a lot of heights based on specific padding + specific line height)

packages/component-library-react/src/Calendar/CalendarNavigationLabel.tsx
```diff
@@ -1,10 +1,9 @@
...
-  label?: string;
-}
+type CalendarNavigationLabelProps = TimeHTMLAttributes<HTMLTimeElement>;
 
-export const CalendarNavigationLabel: FC<CalendarNavigationLabelProps> = ({ label }) => (
-  <Heading2 className="utrecht-calendar__navigation-label">{label}</Heading2>
+export const CalendarNavigationLabel: FC<CalendarNavigationLabelProps> = ({ children, ...props }) => (
+  <time className="utrecht-calendar__navigation-label" {...props}>
+    {children}
+  </time>
```
bddjong, line 8: Taken from Den Haag datepicker, semantic element for time with machine readable time property.

Yolijn: Nice I like `<time>` for this! However I think we might also want to include the `datetime` attribute then. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time
It would mean adding an extra property though :)

Yolijn: Oh never mind me, you've added that already! 👍😍

Yolijn (approved):

PR #1700, opened 2023-04-19
feat: index.ts status badge toegevoegd

scar055:


scar055:
in het badge status mapje de indes.scss waren de background colors verkeerd ingesteld bij de statusen

Yolijn:
@Robbert Hmmm is there good reason for these CSS variables to have fallback values anyway? I do agree with @scar055 you would then expect the error to be red... but it doesn't really match our approach from elsewhere to not have any defaults to begin with 🤷

Robbert (approved):
I agree we should probably remove the fallback values at some point, but there is no harm to make them more accessible in the meanwhile.

PR #1702, opened 2023-04-20
chore: fix NX commands to use correct project names

Robbert:
previously, NX generated incorrect project names, and we adapted our commands to use those incorrect names. Now NX has fixed that, and our commands no longer worked.


bddjong (approved):
Tested it, works again as expected :D

PR #1752, opened 2023-05-31
refactor: separate state from calendar rendering

Robbert:
...and parse the date fewer times

### DIFF
--- a/packages/component-library-react/src/Calendar/index.tsx
+++ b/packages/component-library-react/src/Calendar/index.tsx
@@ -37,6 +37,17 @@ import { IconArrowLeftDouble } from './IconArrowLeftDouble';
 import { IconArrowRight } from './IconArrowRight';
 import { IconArrowRightDouble } from './IconArrowRightDouble';
 
+/* date-fns `Day` type */
+enum Day {
+  SUNDAY = 0,
+  MONDAY = 1,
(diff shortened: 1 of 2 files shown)


packages/component-library-react/src/Calendar/index.tsx
```diff
@@ -48,78 +59,144 @@ function createCalendar(today: Date): Date[] {
...
+
+  let defaultDate = typeof defaultValue === 'string' ? parseISO(defaultValue) : undefined;
+
+  if (currentDate instanceof Date) {
+    deprecated('currentDate', 'defaultValue');
+    defaultDate = currentDate;
+  }
+
+  const [visibleMonth, setVisibleMonth] = useState(defaultDate || now);
+  const [selectedDate, setSelectedDate] = useState(defaultDate);
```
sergei-maertens, line 182: I think I also noticed an issue where when the calendar is visible and the `currentDate` prop changes, the state does not update accordingly. It sort of works for us because when you type the date we close the calendar/date picker, but you can resolve this with:

```tsx
if (selectedDate != defaultDate) {
  setSelectedDate(defaultDate);
}
```

Reference for the state update while rendering: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

PR #1764, opened 2023-06-14
ci: fix the install pnpm issue on github actions

AliKdhim87:


Robbert:
We need the lockfile with high accuracy!

PR #1772, opened 2023-06-22
feat: actions slot for alert

Robbert:


Robbert:
The grid layout causes the `gap` to show up between the `icon` area and the `message` area, as well as between the `message` area and the `action_` area. The alert now has too much white space. Should fix this before merging.

PR #1788, opened 2023-08-09
build: output each React component separately too

Robbert:
PR with experiment that follows discussion on Slack: https://codefornl.slack.com/archives/C01DAT4TRPF/p1691568231239299

Would it be possible to import only one component, with only the CSS for that one component?

Also: Would it be possible to build import components without having to install peerDependencies for components you don't use? For example: use `Header` without installing the `peerDependency` `date-fns` for `Calendar`.


Robbert:
The major problem with building all these components appears to be memory usage. Rollup uses more than 15GB of memory on my mac.

I tried to make 26 chunks, one `npm run` for each letter (first build components `A*`, then `B*`, etc.), which solves the problem for my mac.

Build takes almost 7 minutes 🙈

But even this appears to randomly crash the GitHub Actions pipeline

Robbert:
We're going with another approach to this #2340

PR #1820, opened 2023-08-20
chore: migrate button css stories to tsx

Yolijn:
As we cannot easily keep te sub folders using one `.stories` file I propose splitting the button stories into multiple files. To group them for development purposes I used a `button` folder. 

And as I now have this folder anyway and the `docs/*.stories.mdx` files all have the `CSS component` story prefix, I've moved those too.


Robbert:
Duplicate of #2247

PR #1859, opened 2023-08-30
chore: refactor searchbar css component story to tsx

AndreaBusse29:


Robbert:
This migration conflicts with Ali's refactoring of the component, so we'll use that other PR with new features instead

PR #1876, opened 2023-09-11
style: remove `letter-spacing` tokens (except for badge)

Robbert:
The `letter-spacing` were only used with `text-transform: uppercase`, without they should be discouraged.


Ollie-nl:
👍🏻

Ollie-nl (approved):

PR #1909, opened 2023-09-23
style: use default focus style when no design tokens are configured

Robbert:


Robbert:
Not sure this works better than `revert`

PR #1923, opened 2023-09-28
style: apply `currentColor` to `<svg>` slotted in `<utrecht-icon>`

Robbert:


Robbert:
Some icons rely on `stroke` instead of `fill`, this would break those icons.

Issue #1887, opened 2023-09-16
Vega dependencies break build


sergei-maertens:
For the package @utrecht/component-library-react, I am using the form components (`FormLabel`, `FormDescription` etc.) with TypeScript. This breaks the build of my library when the vega dependencies are not present (see #1784 for the PR that added these).

Environment information:

```
➜  formio-renderer git:(feature/mvp-renderer) ✗ npm list @utrecht/component-library-react
@open-formulieren/formio-renderer@0.1.0 /home/bbt/code/formio-renderer
└── @utrecht/component-library-react@1.0.0-alpha.366
```

Build error:

```
➜  formio-renderer git:(feature/mvp-renderer) ✗ npm run build:esm

> @open-formulieren/formio-renderer@0.1.0 build:esm
> tsc -p tsconfig.prod.json && tsc-alias -p tsconfig.prod.json

node_modules/@utrecht/component-library-react/dist/VegaVisualization.d.ts:5:27 - error TS2307: Cannot find module 'react-vega/lib/Vega' or its corresponding type declarations.

5 import { VegaProps } from 'react-vega/lib/Vega';
                            ~~~~~~~~~~~~~~~~~~~~~

node_modules/@utrecht/component-library-react/dist/VegaVisualization.d.ts:6:24 - error TS2307: Cannot find module 'vega' or its corresponding type declarations.

6 import { Config } from 'vega';
                         ~~~~~~


Found 2 errors in the same file, starting at: node_modules/@utrecht/component-library-react/dist/VegaVisualization.d.ts:5
```

To resolve this, I needed to add these to my devDependencies:

```
"react-vega": "^7.6.0",
"vega": "^5.25.0",
"vega-lite": "^5.15.0"
```

Probably they should be in the `peerDependencies` of the package?

Robbert:
@sergei-maertens Thanks! We will adjust the build script to avoid this issue.

Robbert:
Fixed! (I hope)

PR #1933, opened 2023-10-01
build: output `Calendar` and `VegaVisualization` separately

Robbert:


Robbert:
`import` paths for `Calendar` and `VegaVisualization` should be updated — but hopefully this fixes #1887.

PR #2005, opened 2023-11-03
feat(css): add `.utrecht-table__caption` css class

AliKdhim87:


components/table/html/_mixin.scss
```diff
@@ -17,6 +17,10 @@
     @include utrecht-table__caption;
   }
 
+  figcaption {
+    @include utrecht-table__figcaption;
+  }
+
```
Robbert, line 23: This would style every `figcaption` this way, that would be undesirable.

So I think we cannot support `figcaption` for tables specifically in the HTML components. My suggestion is to remove this code.

packages/storybook-css/src/Table.stories.tsx
```diff
@@ -1910,7 +1910,9 @@ export const FigureCaption: Story = {
   render: (({ descriptionId, children }: PropsWithChildren<{ descriptionId: string }>) => (
     <figure>
       <Table aria-describedby={descriptionId}>{children}</Table>
-      <figcaption id={descriptionId}>Bron: Gemeente Utrecht</figcaption>
+      <figcaption className="utrecht-table__figcaption" id={descriptionId}>
+        Bron: Gemeente Utrecht
+      </figcaption>
```
Robbert, line 1915: I would want to make the `figcaption` optional and make the text "Bron: Gemeente Utrecht" configurable from the properties.

Robbert (changes requested):

Robbert (approved):

PR #2135, opened 2024-02-06
build: update package.json#engines

matijs:
Use the current LTS major version of NodeJS
Use the current major version of pnpm


Robbert (changes requested):
.nvmrc update is missing

updating `@types/node` also seems reasonable

matijs:
> .nvmrc update is missing
> 
> updating `@types/node` also seems reasonable

Done and done.

Robbert (approved):

PR #2184, opened 2024-03-19
fix(react-library): add the missing attribute to the accordion panel

AliKdhim87:


packages/component-library-react/src/Accordion.tsx
```diff
@@ -115,7 +115,9 @@ export const AccordionSection = forwardRef(
             {children}
           </section>
         ) : (
-          <div {...panelAttributes}>{children}</div>
```
Robbert, line 118: I wonder if `aria-labelledby={buttonId}` works in this case, because I think you can only label things with a (implied) `role`.

Robbert: The `section` role is optional to avoid "landmark overload", so no label for the div is intentional in this case. I don't think this needs fixing.

Adding `id={panelId}` is a good improvement of course.

savitris: I also believe the aria-labelledby can go. 
It could indeed lead to a very busy landmark list rotor.

savitris (changes requested):

Robbert (approved):

PR #2212, opened 2024-05-17
style: add `min-block-size` to select and make the default minimum 44px

Robbert:
This is a breaking change for folks who have select components smaller than 44px, should I be more careful?

Het gaat om deze stories:

- [CSS Storybook — Select](https://utrecht-git-style-select-min-block-size-nl-design-system.vercel.app/storybook-css/index.html?path=/docs/css-select--docs)
- [CSS Storybook — Select design tokens](https://utrecht-git-style-select-min-block-size-nl-design-system.vercel.app/storybook-css/index.html?path=/story/css-select--design-tokens)


Robbert:
Hm... misschien moeten we ook nog `line-height` toevoegen nu.

bddjong (approved):

PR #2215, opened 2024-05-17
build: update actions

matijs:
- update actions to their latest versions
- make action versions explicit
- re-order "pnpm/action-setup" and "actions/setup-node" so "actions/cache" is no longer needed

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -16,34 +16,20 @@ jobs:
 
     steps:
       - name: Checkout branch
-        uses: actions/checkout@v4
+        uses: actions/checkout@v4.1.6
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4.0.0
 
       - name: Set up Node.js version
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-delivery.yml
```diff
@@ -54,34 +40,20 @@ jobs:
...
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
+          node-version-file: .nvmrc
+          cache: pnpm
 
       - name: Install dependencies
         run: |
-          pnpm install --frozen-lockfile
+          pnpm install
```
Robbert, line 56: @matijs I kind of like having an explicit `--frozen-lockfile` instead of relying on `CI=1` environment variable magic. Any particular reason to remove it?

matijs: We can add it again. The only reason I removed it was that is was free and in line with the other repos.

PR #2230, opened 2024-05-31
feat: add Blockquote component for React

Robbert:
[Blockquote in Storybook React](https://utrecht-git-feat-blockquote-react-nl-design-system.vercel.app/storybook-react/?path=%2Fstory%2Freact-blockquote--default)

### DIFF
--- a/packages/component-library-react/src/Blockquote.test.tsx
+++ b/packages/component-library-react/src/Blockquote.test.tsx
@@ -1,0 +1,125 @@
+import { render } from '@testing-library/react';
+import { createRef } from 'react';
+import { Blockquote } from './Blockquote';
+import '@testing-library/jest-dom';
+
+describe('Blockquote', () => {
+  it('renders an HTML blockquote element', () => {
+    const { container } = render(<Blockquote />);
+
+    const blockquote = container.querySelector('blockquote:only-child');
+
+    expect(blockquote).toBeInTheDocument();
+  });
+
+  it('renders a block element', () => {
+    const { container } = render(<Blockquote />);
+
+    const paragraph = container.querySelector(':only-child');
+
+    expect(paragraph).toHaveStyle({ display: 'block' });
+  });
+
+  it('renders a design system BEM class name: utrecht-blockquote', () => {
+    const { container } = render(<Blockquote />);
+
+    const paragraph = container.querySelector(':only-child');
+
+    expect(paragraph).toHaveClass('utrecht-blockquote');
+  });
+
+  it('renders rich text content', () => {
+    const { container } = render(
+      <Blockquote>
+        <code>Hello, world</code>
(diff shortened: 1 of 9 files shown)


packages/component-library-react/src/Blockquote.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  it('renders a design system BEM class name: utrecht-blockquote', () => {
+    const { container } = render(<Blockquote />);
+
+    const paragraph = container.querySelector(':only-child');
+
+    expect(paragraph).toHaveClass('utrecht-blockquote');
+  });
+
+  it('renders rich text content', () => {
```
Yolijn, line 31: I would propose checking if it allows for multiple paragraphs as rich text as I think just one might give a false positive in some implementations where the blockquote always contains a paragraph anyway 🤓

packages/component-library-react/src/css-module/Blockquote.tsx
```diff
@@ -0,0 +1,8 @@
+/**
+ * @license EUPL-1.2
+ * Copyright (c) 2021 Robbert Broersma
+ */
+
+import '../../../../components/blockquote/css/index.scss';
```
Yolijn, line 6: I think it might be nice if this was referencing the css-component package instead?

packages/storybook-react/src/stories/Blockquote.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+import React from 'react';
+import { designTokenStory } from './util';
+
+const meta = {
+  title: 'React Component/Blockquote',
+  id: 'react-blockquote',
+  component: Blockquote,
+  tags: ['autodocs'],
+  args: {
+    attribution: undefined,
```
Yolijn, line 15: is it needed to explicitly define `attribution` as undefined here?

packages/storybook-react/src/stories/Blockquote.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+        consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+        pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est
+        laborum
+      </Paragraph>
+    ),
+  },
+  argTypes: {
+    lead: {
+      name: 'lead',
```
Yolijn, line 28: These argTypes are left from a copied Paragraph story I think? I would expect attribution to be mentioned here instead?

Yolijn (approved):

PR #2233, opened 2024-06-02
build: provide pre-built CSS in CSS component packages

Robbert:


components/form-toggle/package.json
```diff
@@ -5,10 +5,21 @@
...
+  "main": "dist/index.css",
+  "scripts": {
+    "build": "rollup -c ../rollup.config.mjs",
+    "clean": "rimraf dist"
+  },
+  "devDependencies": {
+    "@utrecht/focus-ring-css": "workspace:*",
+    "rollup": "3.29.4"
+  },
+  "dependencies": {
```
Yolijn, line 20: I don't think we need the focus-ring-css as a dependency here as it's already a dev-dependency like everywhere else

packages/component-library-react/rollup.config.mjs
```diff
@@ -42,9 +42,8 @@ export default [
         globals: outputGlobals,
       },
     ],
-    external: [/@babel\/runtime/, 'react-dom', 'react'],
     plugins: [
-      peerDepsExternal({ includeDependencies: true }),
+      peerDepsExternal({ includeDependencies: false }),
```
Yolijn, line 46: I propose removal of the `includeDependencies` option as it's no longer needed and it's confusing to have it specifically mentioned.

packages/storybook-css/src/LinkList.tsx
```diff
@@ -9,7 +9,6 @@ import { Link } from '@utrecht/component-library-react';
 import { UtrechtIconArrow } from '@utrecht/web-component-library-react';
 import clsx from 'clsx';
 import React, { HTMLAttributes, PropsWithChildren, ReactNode } from 'react';
-import '@utrecht/components/link-list/css/index.scss';
```
Yolijn, line 12: Hmm interesting I would expect this CSS to be needed still? 

However this story should probably be refactored anyway and use the React LinkList component. 🤔

Yolijn (approved):

PR #2241, opened 2024-06-11
style: add link-list.row-gap token and other link list tweaks

Robbert:
This is part of getting the [Link list in the NL Design System community Figma](https://www.figma.com/design/gX3fq7k0uUKC45AJbuItDv/Local---Voorbeeld---Bibliotheek?node-id=1548-6863&t=KuOvWHVyjBq0moFp-4) consistent with the code from the Utrecht community component.

See also: [Link list GitHub discussion at NL Design System](https://github.com/orgs/nl-design-system/discussions/266)

### DIFF
--- a/components/link-list/src/_mixin.scss
+++ b/components/link-list/src/_mixin.scss
@@ -17,10 +17,6 @@
 }
 
 @mixin utrecht-link-list {
-  --utrecht-link-text-decoration: none;
-  --utrecht-link-hover-text-decoration: underline;
-  --utrecht-link-focus-text-decoration: underline;
-
   display: flex;
   flex-direction: column;
   margin-block-end: calc(var(--utrecht-space-around, 0) * var(--utrecht-link-list-margin-block-end, 0));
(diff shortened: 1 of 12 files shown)


components/link-list/src/_mixin.scss
```diff
@@ -0,0 +1,44 @@
...
+  --utrecht-link-icon-size: var(--utrecht-link-list-icon-size);
+  --utrecht-icon-inset-block-start: var(--utrecht-link-list-icon-inset-block-start);
+
+  display: block;
+  font-weight: var(--utrecht-link-list-item-font-weight);
+}
+
+@mixin utrecht-link-list__link {
+  align-items: baseline;
+  column-gap: var(--utrecht-link-list-item-column-gap);
```
Yolijn, line 42: Should this be `--utrecht-link-list-link-column-gap` or should `column-gap` be set in the mixin for the `__item` instead?

Yolijn: Also I'm wondering what the difference is between een link-list__item and a link-list__link... the React component also only implements the `LinkListLink`?

Robbert: you make some very good points

Robbert: done

Yolijn (approved):

PR #2248, opened 2024-06-23
feat: support indeterminate in Checkbox for React

Robbert:


packages/component-library-react/src/Checkbox.tsx
```diff
@@ -1,36 +1,64 @@
...
   invalid?: boolean;
 }
 
 export const Checkbox = forwardRef(
   (
-    { appearance = 'custom', disabled, invalid, required, className, ...restProps }: CheckboxProps,
+    {
+      appearance = 'custom',
+      disabled,
+      indeterminate = false,
```
Yolijn, line 15: I wonder if `undefined` wouldn't already be sufficient? Is an actual `false` value required for the indeterminate class name to work?

Yolijn (approved):

PR #2282, opened 2024-07-04
feat: :wheelchair: feat(added aria-hidden="true" to svg icons for bet…

MeesoPost:
### add aria-hidden="true"  to [Utrecht SVG icon set](https://github.com/nl-design-system/utrecht/tree/main/components/icon) ♿️

- [ ]  The element is hidden from the accessibility API. 

I added the attribute to all the SVG icons by adding it to the svgo.config.mjs file so all SVG icons are affected with
the change
with the added change i tested it in the Utrecht storybook and the icon is hidden/ignored, but still visible on the page

image

### DIFF
--- a/components/icon/svgo.config.mjs
+++ b/components/icon/svgo.config.mjs
@@ -29,6 +29,7 @@ export default {
             fill: 'currentColor',
             height: '1em',
             width: '1em',
+            'aria-hidden': 'true',
           },
         ],
       },


Robbert:
@MeesoPost Zou je ook de icon component willen controleren in alle framework storybooks? Bijvoorbeeld: misschien moet `aria-hidden` ook nog toegevoegd worden aan de `<utrecht-icon>` web component en de `<Icon/>` React component.

PR #2293, opened 2024-07-11
style: restore font-smoothing override

Robbert:


components/document/src/_mixin.scss
```diff
@@ -3,8 +3,15 @@
  * Copyright (c) 2021 Robbert Broersma
  */
 
-@mixin utrecht-document {
+@mixin reset-font-smoothing() {
   /* reset `font-smoothing: antialiasing`, prefer automatic (`subpixel-antialiasing`) behavior for high-dpi screens */
```
Robbert, line 7: De comment was er nog wel, maar de code niet meer 🤔

Hiermee is de volgende code niet meer nodig in de NL Design System website, hoop ik:
https://github.com/nl-design-system/documentatie/commit/7e3e2622199fcd244def476a89c64b0b85050392

components/document/src/_mixin.scss
```diff
@@ -3,8 +3,15 @@
  * Copyright (c) 2021 Robbert Broersma
  */
 
-@mixin utrecht-document {
+@mixin reset-font-smoothing() {
```
Yolijn, line 6: I don't think we need the `()` after `reset-font-smoothing` we do not use it anywhere else :)

Yolijn (approved):

PR #2367, opened 2024-08-19
fix: metadata for backdrop

jeffreylauwers:
Fixed metadata for backdrop.

[Available tokens in Token Studio](https://docs.tokens.studio/available-tokens/available-tokens)

Not supported
- `z-index`
- `animation-duration`

Supported
- `opacity`


jeffreylauwers:
Did not work. Will try again.

PR #2420, opened 2024-08-23
feat: add metadata for mapcontrolbutton

jeffreylauwers:
- Added metadata for mapcontrolbutton tokens.
- Added missing tokens for variables that were in the index.scss file.
- Removed the tokens `mapcontrolbutton.disabled.hover-background-color` and `mapcontrolbutton.primary-action.hover-background-color` as I assume it was a mistake.
- Could not find any mention of `mapcontrolbutton.primary-action*` tokens but kept them in there.
- Also wonder about the naming. Should it be mapcontrol-button? Anyway... kept it this way.

### DIFF
--- a/.changeset/sheet-shock-round.md
+++ b/.changeset/sheet-shock-round.md
@@ -1,5 +1,5 @@
 ---
-"@utrecht/mapcontrolbutton}-css": major
+"@utrecht/map-control-button-css": major
 ---
 
 - Added metadata for `mapcontrolbutton` tokens.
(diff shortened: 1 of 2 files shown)


.changeset/sheet-shock-round.md
```diff
@@ -0,0 +1,7 @@
+---
+"@utrecht/mapcontrolbutton}-css": major
```
Robbert, line 2:
```suggestion
"@utrecht/map-control-button-css": major
```

Robbert (changes requested):

.changeset/sheet-shock-round.md
jeffreylauwers: Ah te veel afgegaan op de naam van het mapje in VS code :)

Robbert (approved):

PR #2443, opened 2024-08-28
fix: metadata for button component

jeffreylauwers:
Changed value of `"nl.nldesignsystem.figma.supports-token"` from `true` to `false` as it's not used for the Figma component within the 'Voorbeeld' library.

### DIFF
--- a/.changeset/scrap-meal-heal.md
+++ b/.changeset/scrap-meal-heal.md
@@ -1,0 +1,5 @@
+---
+"@utrecht/button-css": minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/scrap-meal-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@utrecht/button-css": minor
```
Robbert, line 2:
```suggestion
"@utrecht/button-css": patch
```

Robbert (changes requested):

Robbert (approved):

PR #2450, opened 2024-08-29
feat: add metadata for select

jeffreylauwers:
- Added metadata for select
- Changed syntax for background-image token

### DIFF
--- a/components/select/src/tokens.json
+++ b/components/select/src/tokens.json
@@ -7,8 +7,20 @@
             "syntax": "<color>",
             "inherits": true
           },
-          "nl.nldesignsystem.fallback": ["utrecht.form-control.background-color"]
-        }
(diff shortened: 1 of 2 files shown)


components/select/src/tokens.json
```diff
@@ -34,107 +48,131 @@
...
       },
       "min-block-size": {
         "$extensions": {
           "nl.nldesignsystem.css.property": {
             "syntax": "<length>",
             "inherits": true
           },
-          "nl.nldesignsystem.fallback": ["utrecht.pointertarget.min-size"]
-        }
+          "nl.nldesignsystem.fallback": ["utrecht.pointertarget.min-size"],
```
Robbert, line 117:
```suggestion
          "nl.nldesignsystem.fallback": ["utrecht.pointer-target.min-size"],
```

Robbert (changes requested):

PR #2552, opened 2024-10-09
feat(component-library-react/paragraph): deprecate lead and small for appearance attribute

### LINKED ISSUE #1800: Consider changing paragraph `lead` boolean attribute to appearance="lead"
`<utrecht-paragraph lead small>` would not work, these are conflicting. Perhaps those two should not be boolean attributes, but one enum.

So instead of:

- `<utrecht-paragraph lead>`
- `<utrecht-paragraph small>`

We'd have:

- `<utrecht-paragraph appearance="lead">`
- `<utrecht-paragraph appearance="small">`

In theory `appearance` could be implemented as a token list in the future (just like the "class" attribute), to allow for `<utrecht-paragraph appearance="small warning">`, then we would be back to square one.

remypar5:
Deprecate the `lead` and `small` attributes in favor of the new `appearance="lead|small"` attribute. They are still supported and don't break any current implementation. When the `appearance` attribute is there, it will take priority.

## Examples

```tsx
<UtrechtParagraph lead>I am a lead paragraph, as I've always been</UtrechtParagraph>
<UtrechtParagraph small>I am a small paragraph, as I've always been</UtrechtParagraph>


<UtrechtParagraph lead appearance="small">
  I am a small paragraph, as the appearance attribute tells me to. I ignore the lead attribute
</UtrechtParagraph>
```

Closes #1800


Robbert:
@remypar5 Thanks for your PR!

remypar5:
In an offline conversation, @Robbert mentioned I didn't add a `changeset`, yet. 

In that same conversation, @matijs mentioned I should remove the unused `default` appearance.

packages/components-react/tsconfig.base.json
```diff
@@ -11,7 +11,7 @@
     "jsx": "react-jsx",
     "lib": ["DOM", "ES2022"],
     "module": "ESNext",
-    "moduleResolution": "Bundler",
+    "moduleResolution": "NodeNext",
```
matijs, line 14: `"module": "ESNext"` + `"moduleResolution": "NodeNext"` is not a valid combination. Did you have an issue that was solved by changing this?

remypar5: Yes, I couldn't run `pnpm run build` nor `pnpm run test`

matijs: What error did you get? `Bundler` was added in TS5 so if your installed versions of TS were stuck on 4, that could explain things.

remypar5: I changed it back. I probably forgot to `pnpm i -f` beforehand. Running into different issues now, though. The `@utrecht/calendar-react` package cannot be found. Not even when I explicitly (manually) build it. It is in the dependencies, so it shouldn't be an issue. 

Just keep digging

Robbert (approved):

PR #2561, opened 2024-10-11
feat: allow FormField's error message position to be changed

remypar5:
This change doesn't change anything it the current behaviour. It just adds the possibility to set the position of the error message (if present) to a certain place in regards to the field which has the faulty value. Either `before` (above) or `after` (underneath). The latter is the default value, representing the current behaviour.


remypar5:
Closing this due to it being stale

PR #2806, opened 2025-04-24
fix: token usage textarea read only variant

dineshduggal:
Currently incorrect tokens are applied, see screenshot:

image

Fixed an issue where the first CSS variable for the `background-color` and `border-color` was not applied due to an incorrect variable name. As a result, the browser was falling back to the next defined variables. This has now been corrected by fixing the variable name, and the intended values for `background-color` and `border-color` are applied properly.

### DIFF
--- a/.changeset/twenty-lizards-fix.md
+++ b/.changeset/twenty-lizards-fix.md
@@ -1,0 +1,5 @@
+---
+"@utrecht/textarea-css": patch
+---
+
+Fix incorrect CSS variable name for Textbox (read-only variant)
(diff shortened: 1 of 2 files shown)


Robbert (commented):
Thanks! Fix looks great, but unless you include a changeset file it will not be released as new npm package version 😅 If you would be so kind as to include that changeset file :)

Documentation: https://nldesignsystem.nl/handboek/developer/changesets

dineshduggal:
> Thanks! Fix looks great, but unless you include a changeset file it will not be released as new npm package version 😅 If you would be so kind as to include that changeset file :)
> 
> Documentation: https://nldesignsystem.nl/handboek/developer/changesets

✅ Done

Robbert (approved):

Issue #3031, opened 2026-01-28
Listbox option selected pseudo states overwrites are always applied


FinnWard:
## TLDR: 
when you set any of the following tokens it will always be applied to the selected listbox option without having the disabled/hover pseudo state:
`--utrecht-listbox-option-selected-disabled-background-color`
`--utrecht-listbox-option-selected-disabled-color`
`--utrecht-listbox-option-selected-hover-background-color`
`--utrecht-listbox-option-selected-hover-color`

## reproduction steps:
- go to the selected option storybook: https://nl-design-system.github.io/utrecht/storybook/?path=/story/css_css-listbox--selected-option
- set the following token `--utrecht-listbox-option-selected-disabled-background-color: red` on the element.style of the selected `li`
- see that the color changes without disabeling the select option 

<img width="1906" height="801" alt="Image" src="https://github.com/user-attachments/assets/949e5b72-5180-4ae0-87d3-46e0a76b02ac" />

## suggested fix:
update the selected class the following way: 
```css
@mixin utrecht-listbox__option--selected {
  --_utrecht-listbox-option-disabled-background-color: var(--utrecht-listbox-option-selected-disabled-background-color);
  --_utrecht-listbox-option-disabled-color: var(--utrecht-listbox-option-selected-disabled-color);
  --_utrecht-listbox-option-interactive-background-color: var(--utrecht-listbox-option-selected-hover-background-color);
  --_utrecht-listbox-option-interactive-color: var(--utrecht-listbox-option-selected-hover-color);
  --_utrecht-listbox-option-state-background-color: var(--utrecht-listbox-option-selected-background-color);
  --_utrecht-listbox-option-state-color: var(--utrecht-listbox-option-selected-color);
}
```

```css
@mixin utrecht-listbox__option--selected {
  --_utrecht-listbox-option-state-background-color: var(--utrecht-listbox-option-selected-background-color);
  --_utrecht-listbox-option-state-color: var(--utrecht-listbox-option-selected-color);
}
```

The overwrite styles will still be applied through the individual classes instead of also being applied through the selected class.

## Regression update:
if you set the above tokens this bug will be flagged in the regression tests

PR #3032, opened 2026-01-28
fix: updated tokens to properly apply the listbox option selected pseudo states

FinnWard:
fixes: https://github.com/nl-design-system/utrecht/issues/3031

I didn't update the figma tokens as this has a larger impact but this does mean that the regression tests wont pick this bug up in case it sneaks back in.

### DIFF
--- a/.changeset/fuzzy-camels-attend.md
+++ b/.changeset/fuzzy-camels-attend.md
@@ -1,5 +1,5 @@
 ---
-"@utrecht/listbox-css": patch
+"@utrecht/listbox-css": minor
 ---
 
 bugfix: updated tokens to properly apply the listbox option selected pseudo states.


.changeset/fuzzy-camels-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@utrecht/listbox-css": patch
```
eslook: I would maybe consider it a minor change instead of a patch change, since people update their package without looking when it's a patch: the fix will cause a visual change which isn't expected when updating a patch.

Robbert:
@FinnWard Bedankt voor je hulp met bugfixes! Ik zie "3 changes must be accepted as baselines" in Chromatic, en ik maak me zorgen dat niet alle effecten gewenst zijn. Zou je willen checken of naar jouw idee alles klopt in Chromatic?

FinnWard:
@eslook en @Robbert, nu met verwachte changes en de changeset geupdate

Issue #2939, opened 2025-10-22
Calendar component: add possiblity to not show weekend days and previous/next year controls


viktorvanwijk:
For an Open Formulieren client, we are making some changes to our appointments form, which contains a datepicker that uses the calendar component. They have requested two changes to the calendar component:

1. Not show weekend days
2. Not show previous/next year controls

Looking at the component props in the calendar story, it currently doesn't seem to be possible to do these things. Would it be OK if we contribute this, or are there any reasons/difficulties to not do this?

PR #3055, opened 2026-02-19
feat(calendar-react): add functionality for not showing weekends and year navigation buttons

vaszig:
Related issue: https://github.com/nl-design-system/utrecht/issues/2939
Closes open-formulieren/open-forms#5692 partly


packages/components-react/calendar-react/src/index.tsx
```diff
@@ -81,6 +82,8 @@ export interface CalendarProps {
   nextMonthButtonTitle?: string;
   minDate?: Date;
   maxDate?: Date;
+  showWeekends: boolean;
+  hideYearControls: boolean;
```
Robbert: Could you make give the properties a consistent API, with naming based on web standards, and preferably singular. For example:

- `displayWeekend`
- `displayYearNavigation`

PR #3125, opened 2026-04-21
refactor: use different approach for textbox container

Robbert:
- span renders border that looks like a Textbox
- border supports forced colors mode for default, focus an disabled
- the icon slots are clickable without JavaScript
- container is `inline-*` like `<input>`, not `display: block`
- focus-visible state is only displayed when the `<input>` has focus, not when any actions have focus
- there are slots that don't change the accessible name: iconStart, iconEnd, actionsStart, actionsEnd
- there are slots nudge folks to choose if something should change the accessible label or description

TODO, for another time perhaps:
Investigate if there should be a `<FormField>` that renders a fieldset but still looks the same, there is a clear group for screenreader users for when there is a combination of `<input>` + `<button>` action, and for `<input>` + `<select>` action.


Robbert:
Replaced by #3130

PR #3129, opened 2026-04-22
feat: add modifiers to set the total amount of columns of the grid

pixelgitter:
Adding CSS modifiers to add the option of setting the total columns of a grid.

## Important Limitations:
- Cannot be set for specific breakpoints

## Use Cases:
- Dashboard layouts with 5 equal items
- Toptasks like https://www.utrecht.nl/zorg-en-onderwijs


pixelgitter:
Replaced by https://github.com/nl-design-system/utrecht/pull/3147

PR #3137, opened 2026-04-27
build: upgrade to Storybook 10

Robbert:
Upgrade everything except Storybook Angular, because that would require updating Angular first

Upgrading Angular is part of a separate PR https://github.com/nl-design-system/utrecht/pull/3138


packages/storybook-html/config/main.ts
```diff
@@ -1,6 +1,11 @@
-import type { StorybookConfig } from '@storybook/react-webpack5';
-import { dirname, join } from 'node:path';
-import * as webpack from 'webpack';
+import type { StorybookConfig } from '@storybook/react-vite';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import { dirname, join, resolve } from 'path';
```
Yolijn:
```suggestion
import { dirname, join, resolve } from 'node:path';
```

packages/storybook-html/config/preview.tsx
```diff
@@ -11,7 +11,7 @@ import React, { ReactElement } from 'react';
 import * as ReactDOMServer from 'react-dom/server';
 
 import 'firacode/distr/fira_code.css';
-import '@utrecht/component-library-css/src/index.scss';
+// import '@utrecht/component-library-css/src/index.scss';
```
Yolijn:
```suggestion
```

Yolijn (approved):

PR #3170, opened 2026-06-15
fix: apply nesting-level modifier classes to React UnorderedList

### LINKED ISSUE #3168: Unordered lists  Bug: list-style-type token is not used
In Open Forms we use react component `<UnorderedList>` and related design tokens. I want to change the default value of the list-stype-type for level 1 from disc to "-". but it's not changed. Other design tokens work as expected. 

Versions I use: 
```
"@utrecht/component-library-react": "14.0.0",
"@utrecht/design-tokens": "^2.5.0",
```

For open forms theme I set up following design tokens:
```
{
  "utrecht": {
    "unordered-list": {
      "font-size": {},
      "line-height": {},
      "margin-block-start": {"value": "12px"},
      "margin-block-end": {"value": "20px"},
      "padding-inline-start": {},
      "item": {
        "margin-block-start": {"value": "4px"},
        "margin-block-end": {},
        "padding-inline-start": {"value": "8px"}
      },
      "marker": {
        "color": {}
      },
      "level-1": {
        "list-style-type": {"value":  "\"–\""}
      }
    }
  }
}
```
All tokens work expect of:
```
 "level-1": {
     "list-style-type": {"value":  "\"–\""}
   }
```

Here is the HTML of the element rendered on the page:
```
<ul role="list" class="utrecht-unordered-list utrecht-unordered-list--level-1">
   <li class="utrecht-unordered-list__item">value A</li>
   <li class="utrecht-unordered-list__item">value B</li>
</ul>
```

The expected result: the markers would be "-" 
The actual result: the markers have default `disc` type 

Here is what I see in the browser

Image

**Upd.:** 
I don't see `level-1` styles  in the "inspect", but I can see them in the CSS
Inspect:

Image

CSS part:
```
.utrecht-unordered-list--html-content,.utrecht-unordered-list--html-content ol ul,.utrecht-unordered-list--level-1{
	--_utrecht-unordered-list-list-style-type: var( --utrecht-unordered-list-level-1-list-style-type, var(--utrecht-unordered-list-list-style-type, ●) );
	--_utrecht-unordered-list-marker-font-size: var( --utrecht-unordered-list-level-1-marker-font-size, var(--utrecht-unordered-list-marker-font-size, 1em) )
```

pixelgitter:
The React `UnorderedList` component never applied `utrecht-unordered-list--level-{n}` classes, so level-specific design tokens (marker style, marker font size) had no effect on the outer list or any nested lists.

Track nesting depth via a React context that increments for each nested `UnorderedList`, and apply the corresponding level class (clamped to level 8, the deepest level defined by the CSS). Add Storybook stories covering nested lists, mixed ordered/unordered nesting, and custom per-level markers.

Closes #3168

### DIFF
--- a/.changeset/silly-otters-march.md
+++ b/.changeset/silly-otters-march.md
@@ -1,0 +1,5 @@
+---
+"@utrecht/component-library-react": patch
+---
+
+Fix `UnorderedList` to apply `utrecht-unordered-list--level-{n}` modifier classes based on nesting depth, so that level-specific design tokens (such as marker style and font size) are applied correctly for nested lists (levels 1–8).
--- a/packages/component-library-react/src/UnorderedList.test.tsx
+++ b/packages/component-library-react/src/UnorderedList.test.tsx
@@ -29,6 +29,14 @@ describe('Unordered list', () => {
     expect(list).toHaveClass('utrecht-unordered-list');
   });
 
+  it('renders a level-1 class name', () => {
+    const { container } = render(<UnorderedList />);
+
+    const list = container.querySelector(':only-child');
+
+    expect(list).toHaveClass('utrecht-unordered-list--level-1');
+  });
+
   it('can contain a list item', () => {
     const { container } = render(
       <UnorderedList>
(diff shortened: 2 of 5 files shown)


sergei-maertens (commented):
I'm not sure if this automagic approach is the best - we might want to override that with a prop if we have other components above or below that use the same class names but aren't aware of the react component hierarchy.

This can lead to nested levels producing the same depth modifiers, or there being gaps between levels.

The bug we reported seems to be something else (I'm investigating), because we're explicitly passing the `className` anyway - and that's fine for us.

PR #3171, opened 2026-06-15
refactor: generate unordered-list level selectors with a Sass loop

pixelgitter:
Replace the 8 hand-duplicated `--level-N` mixins and selector chains in `_mixin.scss`, `index.scss`, and `html/_mixin.scss` with a shared `utrecht-unordered-list-nesting()` function and `@for` loops. Compiled CSS output is unchanged.


pixelgitter:
@sergei-maertens noted that the readability of the old situation is better, which is a fair assessment. This PR might not be as helpful as initially thought. If this conclusion remains, the PR will be closed.

PR #3174, opened 2026-06-15
fix(accordion): apply accordion button icon size token

BartWaardenburg:
The `utrecht.accordion.button.icon.size` token already exists (24px, `figma-implementation: true`) but was never applied, so the icon kept its intrinsic size and the panel text couldn't be lined up with the button label. This applies it on `.utrecht-accordion__button-icon` as the icon column width, and exposes `--utrecht-icon-size` to icon children.

Heads up for review: the default chevron in `Accordion.tsx` is a fixed 14x8 SVG, so it doesn't scale with `--utrecht-icon-size`; this mainly gives the icon a predictable column width. Whether to also center the chevron in that column, or make the default SVG scalable, is worth a quick look. I marked the changeset `minor` since it changes the default rendering.

### DIFF
--- a/.changeset/accordion-button-icon-size.md
+++ b/.changeset/accordion-button-icon-size.md
@@ -2,4 +2,6 @@
 "@utrecht/accordion-css": minor
 ---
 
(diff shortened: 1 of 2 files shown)


FinnWard (commented):
de changes zien er goed uit maar de chromium changes zijn wat appart. de rtl story werkt zoals ik verrwacht maar de anderen  krijgen niet de uitlijning die ik verwacht https://www.chromatic.com/test?appId=62d564c42f1b71308b19d073&id=6a300a36ddfa881f3ed71fad

FinnWard (approved):
lekker bezig!

Robbert (approved):

PR #3183, opened 2026-07-02
feat(textbox): add hover state tokens

BartWaardenburg:
Adds `utrecht.textbox.hover.{background-color,border-color,color}` and applies them on `:hover`, placed before the other states in source order so focus, invalid, read-only and disabled keep precedence. The fallback chain includes the shared `--utrecht-form-control-hover-*` layer that the Select already uses, so a theme can style hover for all form controls in one place: `textbox-hover → form-control-hover → textbox → form-control`. The tokens have no default value, so the default appearance is unchanged.

Scope note: #2716 mentions `hover.background-color` and `hover.color`; this also adds `hover.border-color`, since a border color change is the most common hover treatment for form controls.

Part of #2716.

### DIFF
--- a/.changeset/stupid-eyes-drive.md
+++ b/.changeset/stupid-eyes-drive.md
@@ -1,0 +1,6 @@
+---
+"@utrecht/textbox-react": minor
+"@utrecht/textbox-css": minor
+---
+
+Add `utrecht-textbox--hover` class name and mixin.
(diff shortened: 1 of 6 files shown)


Robbert (changes requested):
Ik mis nog een `utrecht-textbox--hover` class name die in Storybook CSS getest kan worden, ik doe zo een commit om 't te fixen

FinnWard (approved):

Robbert (approved):

PR #3184, opened 2026-07-02
feat(textarea): add hover state tokens

BartWaardenburg:
Adds `utrecht.textarea.hover.{background-color,border-color,color}` and applies them on `:hover`, before the other states in source order so focus, invalid, read-only and disabled keep precedence. Same approach as the Textbox equivalent, including the shared `--utrecht-form-control-hover-*` fallback layer that the Select already uses: `textarea-hover → form-control-hover → textarea → form-control`. The tokens have no default value, so the default appearance is unchanged.

Scope note: #2716 mentions `hover.background-color` and `hover.color`; this also adds `hover.border-color`.

Part of #2716.

### DIFF
--- a/.changeset/gentle-panthers-accept.md
+++ b/.changeset/gentle-panthers-accept.md
@@ -1,0 +1,6 @@
+---
+"@utrecht/textarea-react": minor
+"@utrecht/textarea-css": minor
+---
+
+Add `utrecht-textarea--hover` mixin and class name.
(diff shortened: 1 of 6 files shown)


Robbert (changes requested):
Ik mis nog een `utrecht-textarea--hover` class name die in Storybook CSS getest kan worden, ik doe zo een commit om 't te fixen

Robbert (approved):

Robbert (approved):

PR #3186, opened 2026-07-03
feat(form-field): add inputId prop to FormFieldTextbox and FormFieldTextarea

BartWaardenburg:
`FormFieldTextbox` and `FormFieldTextarea` always generate the id of the form control with `useId()`, so consumers cannot reference the input from elsewhere in the document. An `id` passed as prop ends up on the form field root element via the props spread, not on the input. This blocks the error summary pattern, where a list of links at the top of the form points at each invalid input (`href="#subject"`), and other patterns that need a predictable id on the control.

This adds an optional `inputId` prop that overrides the generated id of the control. The `<label for>` and `aria-describedby` wiring keep working, and the default behavior is unchanged when the prop is not set. The prop name follows the existing `inputRef` and `inputDir` props; the `id` prop keeps targeting the form field root element.

### DIFF
--- a/.changeset/form-field-input-id.md
+++ b/.changeset/form-field-input-id.md
@@ -1,0 +1,5 @@
+---
+"@utrecht/component-library-react": minor
+---
+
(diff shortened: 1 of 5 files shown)


Robbert (approved):
👨🏻‍🍳🤌

PR #3211, opened 2026-08-07
fix: `pnpm run --recursive init`

Robbert:
fixes the pipeline maybe

https://github.com/nl-design-system/utrecht/actions/runs/31171828699/job/92845380630?pr=3192


Yolijn (approved):
🤞

